PR #1, opened 2019-08-14
feat: allow optional HTTPS for git origin

codebytere:
Allows using HTTPS instead of SSH when resetting Electron's & Node's origins.

cc @MarshallOfSound

### DIFF
--- a/win/commands/sync.bat
+++ b/win/commands/sync.bat
@@ -13,17 +13,17 @@ if %errorlevel%=="1" goto :fail
 echo Updating git remotes
 
 cd electron
-call git remote set-url origin git@github.com:electron/electron.git
+call git remote set-url origin %ELECTRON_GIT_ORIGIN%
 if %errorlevel%=="1" goto :fail
 
-call git remote set-url origin --push git@github.com:electron/electron.git
+call git remote set-url origin --push %ELECTRON_GIT_ORIGIN%
 if %errorlevel%=="1" goto :fail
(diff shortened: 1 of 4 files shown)


win/commands/sync.bat
```diff
@@ -13,17 +13,17 @@ if %errorlevel%=="1" goto :fail
 echo Updating git remotes
 
 cd electron
-call git remote set-url origin git@github.com:electron/electron.git
+call git remote set-url origin $ELECTRON_GIT_ORIGIN
```
MarshallOfSound:
```suggestion
call git remote set-url origin %ELECTRON_GIT_ORIGIN%
```

Windows env vars are used like that ☝️

codebytere: oh TIL

win/commands/sync.bat
```diff
@@ -13,17 +13,17 @@ if %errorlevel%=="1" goto :fail
 echo Updating git remotes
 
 cd electron
-call git remote set-url origin git@github.com:electron/electron.git
+call git remote set-url origin %ELECTRON_GIT_ORIGIN%
 if %errorlevel%=="1" goto :fail
 
-call git remote set-url origin --push git@github.com:electron/electron.git
+call git remote set-url origin --push $ELECTRON_GIT_ORIGIN
```
MarshallOfSound: Gotta do these other ones too

PR #2, opened 2019-08-15
fix: anonymous functions don't get hoisted

codebytere:
This was failing on:

```sh
electron-gn-scripts on git:master ❯ e generate-config                          7:53PM
/Users/codebytere/electron-gn-scripts/common/validate-env.js:22
  fail('You must create a config.yml file in the root of this repository, copy config.example.yml to get started');
  ^

ReferenceError: Cannot access 'fail' before initialization
    at Object.<anonymous> (/Users/codebytere/electron-gn-scripts/common/validate-env.js:22:3)
.... rest of stacktrace
```

### DIFF
--- a/common/validate-env.js
+++ b/common/validate-env.js
@@ -1,7 +1,11 @@
 const chalk = require('chalk').default;
 const fs = require('fs');
-const path = require('path');
 const yaml = require('yaml-js');
+
+const fail = (message) => {
+  console.error(chalk.red(message));
+  process.exit(1);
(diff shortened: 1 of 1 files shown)


common/validate-env.js
```diff
@@ -1,8 +1,12 @@
 const chalk = require('chalk').default;
 const fs = require('fs');
-const path = require('path');
```
codebytere: this was unused

PR #4, opened 2019-08-31
feat: add command for patch exporting

codebytere:
Adds a new command `e export-patches [dir]` to export patches to the desired patch folder in Electron source tree.

cc @MarshallOfSound

### DIFF
--- a/nix/commands/export-patches.sh
+++ b/nix/commands/export-patches.sh
@@ -1,0 +1,33 @@
+#!/usr/bin/env bash
+
+set -e
+
+basedir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
+
+source $basedir/__load-config.sh
+
+# wrapper fn to check if a key exists in a hashmap
+function array_key_exists() {
+  local _array_name="$1"
+  local _key="$2"
+  local _cmd='echo ${!'$_array_name'[@]}'
+  local _array_keys=($(eval $_cmd))
+  local _key_exists=$(echo " ${_array_keys[@]} " | grep " $_key " &>/dev/null; echo $?)
+  [[ "$_key_exists" = "0" ]] && return 0 || return 1
+}
+
+# init hashmap of valid patch directories
+declare -A patch_dirs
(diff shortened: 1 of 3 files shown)


nix/commands/export-patches.sh
```diff
@@ -0,0 +1,33 @@
...
+
+basedir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
+source $basedir/__load-config.sh
+
+SRC_DIR=$ELECTRON_GN_ROOT/src
+
+if [[ "$(array_key_exists 'patch_dirs' "$@"; echo $?)" = "0" ]]; then
+  PATCH_FOLDER_NAME=$SRC_DIR/${patch_dirs[$@]}
+  cd $PATCH_FOLDER_NAME
+  $SRC_DIR/electron/script/git-export-patches -o $SRC_DIR/electron/patches/"$@"
```
MarshallOfSound: `$@` is "all args" I think we only want $1 here.

nix/commands/export-patches.sh
```diff
@@ -0,0 +1,33 @@
...
+}
+
+# init array of valid patch directories
+declare -A patch_dirs
+patch_dirs["node"]="third_party/electron_node"
+patch_dirs["v8"]="v8"
+patch_dirs["chromium"]=""
+patch_dirs["boringssl"]="third_party/boringssl"
+
+basedir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
```
MarshallOfSound: Just to match the other files can this line and source the `__load-config` file be moved to the top of this one.  Just under `set -e`

nix/commands/export-patches.sh
```diff
@@ -0,0 +1,33 @@
...
+source $basedir/__load-config.sh
+
+SRC_DIR=$ELECTRON_GN_ROOT/src
+
+if [[ "$(array_key_exists 'patch_dirs' "$@"; echo $?)" = "0" ]]; then
```
MarshallOfSound: Same comment here RE `$@` do we ever use more than the $1 value?

PR #11, opened 2019-09-16
feat: add `e show`

ckerr:
I started to submit these as separate standalone PRs, but since this would've generated a bunch of minor PRs and the changes are imo uncontroversial, I thought I'd save some churn by submitting everything in a clearinghouse PR.

Features:
 * add `e src` and `e exec` utils to echo back paths to executables and source directories. Useful for passing to other shell scripts.
 * in `e test`, rebuild the node headers if necessary
 * when `e debug` uses gdb, load Chromium's gdbinit file
 * in `e debug`, pass extra user-provided args along to the debugger invocation
 * add `e make` as an alias to `e build` via symlinks

Fixes:
 * `e pr` uses xdg-open on Linux, open on macOS

Refactors:
 * we calculated the exec path in several places; consolidate that into one place
 * run all the scripts through `shellcheck` linter. tl;dr: double-quote all the things
 * use `$OSTYPE` as a cheaper way to decide whether we're running on Linux or Darwin
 * various log message tweaks, e.g. since `e generate-config` was removed in ffadc2974305fbc372b97c8c1405c1ec2c07ed0d, don't tell people to run it iff config can't be found


README.md
```diff
@@ -210,6 +210,41 @@ Valid patch directories include:
...
+$ e exec
+/home/username/electron-gn-root/src/out/Testing/electron
+
+$ uname
+Darwin
+$ e exec
+/Users/username/electron-gn-root/src/out/Testing/Electron.app/Contents/MacOS/Electron
+```
+
+### `e src`
```
codebytere: maybe we could consolidate these PRs under one umbrella, since they both serve a similar purpose?

perhaps
```
e show [path]
```

ckerr: Mmmaybe? If there are more use cases for an `e show` than just exec + src, I could see that as a logical grouping.

If in practice we only have the two cases, I've got a slight preference to keep them as-is because `e show src base` is borderline prolix.

MarshallOfSound: Just as a third one I had in an old script `e show out` to get the path to the current out directory.  Used in an old alias to run `mksnapshot` (`${egetout}/mksnapshot`).  Not set either way just noting a third potential path people may want

nix/commands/make.sh
```diff
@@ -0,0 +1 @@
+build.sh
```
MarshallOfSound: Is this a symlink or a file? Hard to tell on github UI

nix/commands/test.sh
```diff
@@ -3,9 +3,30 @@
...
+# the build process.
+declare -r build_dir="${ELECTRON_GN_ROOT}/src/out/${ELECTRON_OUT_DIR}"
+declare -r node_headers_dir="${build_dir}/gen/node_headers"
+declare -r electron_spec_dir="${ELECTRON_GN_ROOT}/src/electron/spec"
 
-cd $ELECTRON_GN_ROOT/src/electron
+# does it need to be rebuilt?
+if [ ! -d "${node_headers_dir}" ]; then
+  declare -r node_headers_need_rebuild='yes'
+elif [ "${electron_spec_dir}/package.json" -nt "${node_headers_dir}" ]; then
```
MarshallOfSound: Neat trick, I can't think of a case where this will false-negative but hey, it's better than what we had before (nothing) 😄

nix/commands/test.sh
```diff
@@ -3,9 +3,30 @@
...
+  declare -r node_headers_need_rebuild='yes'
+else
+  declare -r node_headers_need_rebuild='no'
+fi
 
-ELECTRON_OUT_DIR=$ELECTRON_OUT_DIR node ./script/spec-runner.js "$@"
+if [ "x$node_headers_need_rebuild" != 'xno' ]; then
+  ninja -C "${build_dir}" third_party/electron_node:headers
+  # install the test modules with the generated headers
+  (cd "${electron_spec_dir}" && npm i --nodedir="${node_headers_dir}")
```
MarshallOfSound: Is this actually required? The `spec-runner` runs the install and this shouldn't be done with `npm` (it should be done with the `script/yarn` file that the spec runner uses?

nix/commands/make.sh
ckerr: both bat and sh are symlinks

nix/commands/test.sh
ckerr: You might be right, this is recycled from my scripts that predate the spec-runner.  This specific line (and the explanation comment on the line above) might not be needed in `test.sh`.

ckerr: removed in 8cd91a23bc99132941c3873777c03edd2188925e

README.md
ckerr: Added an umbrella subcommand `get` and the outdir option in d08d9e8b0474bd870d9c60082ee36886308e2262

README.md
```diff
@@ -210,6 +210,36 @@ Valid patch directories include:
 
 **Nota Bene:** You need to be running at least Bash v4 to use this command.
 
+### `e get`
```
MarshallOfSound: To avoid confusion with [`electron/get`](https://github.com/electron/get) we should probably name this something else

ckerr: Renamed to `e show` in 307f949

PR #12, opened 2019-09-18
feat: use chalk-like colors on nix

ckerr:
a single-topic PR this time, does what it says in the title:

use chalk-like colors in the nix scripts' console log messages.


ckerr:
let's clean up that history first though.

PR #13, opened 2019-09-18
feat: on nix, add chalklike color-coded messages

ckerr:
does what it says in the title: adds color to the nix log messages similar to our chalk messages in js scripts


codebytere:
@ckerr would you mind adding some screenshots?

ckerr:
@codebytere sure. This is from the automate-depot-tools-and-yarn branch, but the colors are the same. Suggestions / improvements welcomed.

screenshot-color

PR #14, opened 2019-09-18
feat: on nix, automate depot-tools installation and 'yarn install'

ckerr:
When invoking depot tools commands used by our scripts (gn, gclient, ninja), ensure that the depot_tools repo is cloned into `electron-gn-scripts/third_party/depot_tools` and temporarily prepend that directory into the path. Also, give `depot_tools` a pull every fortnight to stay up-to-date.

Likewise, ensure that `yarn install` has been called before invoking node.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,61 +6,39 @@ This repository contains some helper/wrapper scripts to make working with GN eas
 
 ### macOS / Linux
 
-```bash
-git clone https://github.com/MarshallOfSound/electron-gn-scripts.git
-cd electron-gn-scripts
-yarn
-# You could also use `npm install` here
-# You should probably add this to your path in your `.zshrc` or `.bashrc`
-export PATH="$PATH:$(pwd)/nix"
+```sh
+$ git clone https://github.com/MarshallOfSound/electron-gn-scripts.git
+$ cd electron-gn-scripts
+$ export PATH="$PATH:/path/to/electron-gn-scripts/nix"
+# ^ You should probably add this to your `~/.profile`
 ```
 
 ### Windows
 
 ```batch
-git clone https://github.com/MarshallOfSound/electron-gn-scripts.git
-cd electron-gn-scripts
-set PATH="%PATH%;"
+PS> git clone https://github.com/MarshallOfSound/electron-gn-scripts.git
+PS> cd electron-gn-scripts\win
+PS> set PATH="%PATH%;%CD%"
 ```
 
-## Setup
+On Windows, you'll also need to install [`depot_tools`](https://commondatastorage.googleapis.com/chrome-infra-docs/flat/depot_tools/docs/html/depot_tools_tutorial.html#_setting_up) as outlined in the [GN Build Instructions](https://github.com/electron/electron/blob/master/docs/development/build-instructions-gn.md) and summarized here:
 
-This toolset does not yet have the ability to initialize an Electron GN setup from scratch so you'll have to
-do the initial work.  These steps are outlined in the [GN Build Instructions](https://github.com/electron/electron/blob/master/docs/development/build-instructions-gn.md) and summarized below.
-
-## Install Depot Tools
-
-You'll need to install [`depot_tools`](https://commondatastorage.googleapis.com/chrome-infra-docs/flat/depot_tools/docs/html/depot_tools_tutorial.html#_setting_up)  to your system.
-
-**On macOS:**
-
-```sh
-# Ensure you're in your home directory
-cd ~
-git clone https://chromium.googlesource.com/chromium/tools/depot_tools.git
-```
-
-Add `depot_tools` to the end of your PATH (you will probably want to put this in your `~/.bashrc` or `~/.zshrc`).
-
-**On Windows:**
-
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -6,62 +6,40 @@ This repository contains some helper/wrapper scripts to make working with GN eas
...
-git clone https://github.com/MarshallOfSound/electron-gn-scripts.git
-cd electron-gn-scripts
-yarn
-# You could also use `npm install` here
-# You should probably add this to your path in your `.zshrc` or `.bashrc`
-export PATH="$PATH:$(pwd)/nix"
+```sh
+$ git clone https://github.com/MarshallOfSound/electron-gn-scripts.git
+$ cd electron-gn-scripts
+$ export PATH="$PATH:/path/to/electron-gn-scripts/nix"
```
MarshallOfSound: I liked the copy-pastability of the `$(pwd)`.  Is there a reason that was killed?

README.md
```diff
@@ -6,62 +6,40 @@ This repository contains some helper/wrapper scripts to make working with GN eas
...
-git clone https://chromium.googlesource.com/chromium/tools/depot_tools.git
+PS> git clone https://github.com/MarshallOfSound/electron-gn-scripts.git
+PS> cd electron-gn-scripts\win
+PS> set PATH="%PATH%;%CD%"
 ```
 
-Add `depot_tools` to the end of your PATH (you will probably want to put this in your `~/.bashrc` or `~/.zshrc`).
-
-**On Windows:**
+On Windows, you'll also need to install [`depot_tools`](https://commondatastorage.googleapis.com/chrome-infra-docs/flat/depot_tools/docs/html/depot_tools_tutorial.html#_setting_up) as outlined in the [GN Build Instructions](https://github.com/electron/electron/blob/master/docs/development/build-instructions-gn.md) and summarized here:
```
MarshallOfSound: I've heard rumors that the download of that zip file is **exactly** the same as cloning the git repo 🤔 Might be worth trying

README.md
```diff
@@ -119,42 +97,41 @@ Example Usage:
...
 ```sh
 # Default - build Electron itself
-e build
+$ e build
 ```
 
 ```sh
 # Build the Electron binary and generates a dist zip file
-e build electron:dist
+$ e build electron:dist
```
MarshallOfSound: Adding this dollar means the commands can't be copy-pasted right? Not sure if that's worth the visual prettiness

nix/commands/__tools.sh
```diff
@@ -0,0 +1,46 @@
...
+    touch "${DEPOT_TOOLS_PATH}"
+  fi
+}
+
+ensure_node_modules() {
+  # if it's missing, install it
+  local -r top="$(git -C "$(dirname "$(readlink -f "$0")")" rev-parse --show-toplevel)"
+  if [[ ! -d "$top/node_modules" ]]; then
+    echo -e "\n\nRunning '${COLOR_CMD}yarn install${COLOR_OFF}' in '${COLOR_DIR}$top${COLOR_OFF}'"
+    yarn --cwd "$top" install
```
MarshallOfSound: Just because not everyone has `yarn` globally.  We should copy what I did in the electron repo
```suggestion
    npx yarn --cwd "$top" install --frozen-lockfile
```

README.md
```diff
@@ -6,62 +6,40 @@ This repository contains some helper/wrapper scripts to make working with GN eas
...
-2) Add `depot_tools` to the start of your `PATH` (must be ahead of any installs of Python)
-3) From a cmd.exe shell, run the command `gclient` (with **no** arguments)
+ 1. Download the `depot_tools` [bundle](https://storage.googleapis.com/chrome-infra/depot_tools.zip) and extract it to `C:\workspace\depot_tools`
+ 2. Add `depot_tools` to the start of your `PATH` (must be ahead of any installs of Python)
+ 3. From a cmd.exe shell, run the command `gclient` (with **no** arguments)
 
 ## Initial Electron Setup
 
-After you've set up `depot_tools`, you'll only need to use the `e` command to perform initial setup.
+Getting and building Electron only requires the `e` command:
```
codebytere: this might be confusing since you still need to do depot_tools on windows

README.md
ckerr: Yes, the comment about 'you should probably add this to your dotfile' -- adding `PATH="$PATH:$(pwd)/nix"` to a dotfile doesn't help anyone.

I've split it into two separate lines in 54f9f6de73a446ccdf2e787d32968b192afd1175 so that it's still copy-pastable

README.md
ckerr: shell prompts removed 18bab5a27589c844eea2657532ce0f85657fe14a

nix/commands/__tools.sh
ckerr: Nice call. Added bfab0b1610b3120b212eb8542b9691f68000e225

README.md
ckerr: Clarified in 3a8cfba2e9aee653ba797804165dffd8f1ef3539

README.md
ckerr: I'm not going to have a Windows setup today, so that's going to have to wait for a followup PR

PR #18, opened 2019-09-20
feat: allow user-specified config file directory

ckerr:
tl;dr: if the environment variable ELECTRON_BUILD_TOOLS_CONFIG exists, the scripts will look for config.\*.yml and generated.env.\* there instead of the default `__dirname, ...` logic.

My goal is to allow these config files to reside in XDG_CONFIG_HOME rather than in the `git clone` directory s.t. my backup scripts can see them, so that they'll persist even if `build-tools/` gets removed and restored, etc.

There are a couple of renames/refactors that make this patch look bigger than it really is:

 * nix/commands/__load-config.sh loads `generated.env.sh`, not a config yaml file, so here the script is renamed to __load-env.sh.

 * each windows batch script hardwired the env loading process, so for DRY I moved this into a new `__load-env.bat` script.

 * common/util.js had a global.OVERRIDE param to let scripts brute-force a specified config file. This was problematic with the custom config directory feature, so here the OVERRIDE has been refactored out and the scripts now take an argument specifying what configFile to use.

 * common/validate-env.js has been renamed load-config.js because it loads and validates a config yaml file, not a generated.env.\* file.

 * common/generate-config.js has been renamed generate-env.js because it generates a generated.env.\* file, not a config yaml file.


common/load-config.js
```diff
@@ -42,16 +35,30 @@ for (const prop of REQUIRED_CONFIG_PROPERTIES) {
...
         } else {
-          fail(`The config property "${prop.name}" must be a path that exists, but "${resolveConfiguredPath(value)}" does not exist :(`);
+          fail(`The config property "${prop.name}" must be a path that exists, but "${resolveConfiguredPath(value)}" does not exist. :(`);
         }
       }
       break;
     }
     default:
-      fail(`The config property "${prop.name}" was configured incorrectly in these helpers scripts, please raise an issue if you see this`);
+      fail(`The config property "${prop.name}" was configured incorrectly in the helpers scripts.  Please raise an issue if you see this.`);
```
codebytere: [nit] extra space here `scripts.  Please raise`

codebytere:
approving assuming this has all been tested locally 😆

PR #20, opened 2019-09-24
chore: add shellcheck linting

codebytere:
What it says on the tin. Add a lint check for bash code. To enable this, i needed to correct some of our bash to align with shellcheck expectations.

cc @MarshallOfSound @ckerr

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,17 @@
+name: Lint
+
+on: [push]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Use Node.js v12.x
+      uses: actions/setup-node@v1
+      with:
+        node-version: 12.x
+    - name: Lint bash code with shellcheck
+      run: |
+        npm install
+        npm run lint
--- a/nix/commands/__tools.sh
+++ b/nix/commands/__tools.sh
@@ -2,12 +2,12 @@
 
 # constants
 
-declare -r -x COLOR_CMD='\033[0;33m' # yellow
-declare -r -x COLOR_DIR='\033[0;36m' # cyan -- same as chalk use in common/
-declare -r -x COLOR_ERR='\033[0;31m' # red
-declare -r -x COLOR_OFF='\033[0m' # no-color
-declare -r -x COLOR_OK='\033[0;32m' # green
-declare -r -x COLOR_WARN='\033[1;31m' # light red
+declare -r COLOR_CMD='\033[0;33m' # yellow
+declare -r COLOR_DIR='\033[0;36m' # cyan -- same as chalk use in common/
+declare -r COLOR_ERR='\033[0;31m' # red
+declare -r COLOR_OFF='\033[0m' # no-color
+declare -r COLOR_OK='\033[0;32m' # green
+declare -r COLOR_WARN='\033[1;31m' # light red
 
 declare __basedir
 __basedir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
@@ -25,7 +25,7 @@ ensure_depot_tools() {
   if [[ ! -d "$DEPOT_TOOLS_PATH" ]]; then
     echo -e "\n\nCloning ${COLOR_CMD}depot_tools${COLOR_OFF} into '${COLOR_DIR}$DEPOT_TOOLS_PATH${COLOR_OFF}'"
     if ! git clone -q 'https://chromium.googlesource.com/chromium/tools/depot_tools.git' "$DEPOT_TOOLS_PATH"; then
-      echo "Failed to clone depot_tools"
+      echo -e "${COLOR_ERR}Failed to clone depot_tools!${COLOR_OFF}"
       exit 1
     fi
   fi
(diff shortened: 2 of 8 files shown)


nix/commands/__load-config.sh
```diff
@@ -11,6 +11,7 @@ fi
 
 source "$config_path"
 
+declare -x ELECTRON_EXEC
```
ckerr: We don't want to export this into the user environment variables -- it's an internal variable used only by the `build-tools` scripts importing it.

Probably best to do something like

```sh
# used by scripts that source this file, so disable "unused variable" warnings
# shellcheck disable=SC2034
declare ELECTRON_EXEC
```

nix/commands/__tools.sh
```diff
@@ -2,15 +2,19 @@
...
-declare -r COLOR_WARN='\033[1;31m' # light red
-
-declare __basedir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
-declare -r ELECTRON_GN_SCRIPTS_ROOT="$(git -C "${__basedir}" rev-parse --show-toplevel)"
+declare -r -x COLOR_CMD='\033[0;33m' # yellow
+declare -r -x COLOR_DIR='\033[0;36m' # cyan -- same as chalk use in common/
+declare -r -x COLOR_ERR='\033[0;31m' # red
+declare -r -x COLOR_OFF='\033[0m' # no-color
+declare -r -x COLOR_OK='\033[0;32m' # green
+declare -r -x COLOR_WARN='\033[1;31m' # light red
```
ckerr: All of the COLOR_ variables are in the same bucket as `ELECTRON_EXEC` and its suggestion applies here. Probably something like

```sh
# used by scripts that source this file, so disable "unused variable" warnings
# shellcheck disable=SC2034
declare -r COLOR_CMD='\033[0;33m' # yellow
# shellcheck disable=SC2034
declare -r COLOR_DIR='\033[0;36m' # cyan -- same as chalk use in common/
...
```

nix/commands/__tools.sh
```diff
@@ -20,16 +24,16 @@ ensure_depot_tools() {
   # if it's missing, install it
   if [[ ! -d "$DEPOT_TOOLS_PATH" ]]; then
     echo -e "\n\nCloning ${COLOR_CMD}depot_tools${COLOR_OFF} into '${COLOR_DIR}$DEPOT_TOOLS_PATH${COLOR_OFF}'"
-    git clone -q 'https://chromium.googlesource.com/chromium/tools/depot_tools.git' "$DEPOT_TOOLS_PATH"
-    if [[ $? -ne 0 ]]; then
-      exit $?
+    if ! git clone -q 'https://chromium.googlesource.com/chromium/tools/depot_tools.git' "$DEPOT_TOOLS_PATH"; then
+      echo "Failed to clone depot_tools"
```
ckerr:
```suggestion
      echo -e "${COLOR_ERR}Failed to clone depot_tools!${COLOR_OFF}"
```

nix/commands/__tools.sh
```diff
@@ -20,16 +24,16 @@ ensure_depot_tools() {
...
+      exit 1
     fi
   fi
 
   # if it's been awhile, update it
   local -r mtime_age_days="$(perl -e 'print int -M $ARGV[0]' "$DEPOT_TOOLS_PATH")"
   local -r update_interval_days=14
-  if (( $mtime_age_days > $update_interval_days )); then
+  if (( mtime_age_days > update_interval_days )); then
```
ckerr: Nice catch on the numerics. Thanks!

nix/commands/__tools.sh
```diff
@@ -40,9 +44,9 @@ ensure_node_modules() {
   # if it's missing, install it
   if [[ ! -d "${ELECTRON_GN_SCRIPTS_ROOT}/node_modules" ]]; then
     echo -e "\n\nRunning '${COLOR_CMD}yarn install${COLOR_OFF}' in '${COLOR_DIR}${ELECTRON_GN_SCRIPTS_ROOT}${COLOR_OFF}'"
-    npx yarn --cwd "${ELECTRON_GN_SCRIPTS_ROOT}" install --frozen-lockfile
-    if [[ $? -ne 0 ]]; then
-      exit $?
+    if ! npx yarn --cwd "${ELECTRON_GN_SCRIPTS_ROOT}" install --frozen-lockfile; then
+      echo "Failed to insall node modules"
```
ckerr:
```suggestion
      echo -e "${COLOR_ERR}Failed to install node modules!${COLOR_OFF}"
```

package.json
```diff
@@ -1,13 +1,19 @@
 {
-  "name": "electron-gn-scripts",
-  "version": "0.0.0",
+  "name": "build-tools",
+  "version": "0.1.0",
   "main": "null",
-  "repository": "https://github.com/MarshallOfSound/electron-gn-scripts",
-  "author": "Samuel Attard <samuel.r.attard@gmail.com>",
+  "scripts": {
+    "lint": "for file in $(find nix -type f); do shellcheck --format=gcc $file; done;"
```
ckerr: 1. Would be better to invoke them in a batch -- it looks like shellcheck _may_ better handle some cases that way, e.g. false "unused variable" warnings for variables declared in one script and used by another.

```suggestion
    "lint": "find nix -type f | xargs shellcheck -f gcc -a"
```

2. Both of these variations require a sh-like environment and `find`, so I think `npm run lint` will fail on Windows? This isn't a blocker at this point, but IDK if there is a better way? :thinking:

.github/workflows/lint.yml
```diff
@@ -0,0 +1,17 @@
+name: Lint
```
ckerr: I really like the idea of adding this to workflows

PR #23, opened 2019-10-06
refactor: port bash, batch files to javascript

ckerr:
Apologies in advance for those who review this; this is basically a rewrite. The goals:

1. Have scripts that actually work everywhere with the same feature set
1. Don't maintain multiple sets of scripts
1. Use a scripting language that all of us are proficient in
1. Add more nvm-like commands to make config management easier.
1. Add cleaner command-line help arguments.
1. Move all the depot-tools invocation logic as possible into the config files so they can be hand-edited as necessary. For example all the `gn gen` logic is created when the config files are created, so is the upstream URL for the electron and electron-node repos (e.g. knowing whether to use git:// or https://). So then the code that calls the depot-tools scripts has no magic logic.
1. Fix some win32 issues, e.g hardcoding `2017` as the GYP_MSYS_VERSION.

Whee, that's quite a lot. And this is a chonky PR, so I'm happy to answer questions / discuss changes.

## Changed Commands

`e init` -- replaces `e fetch`. Command name borrowed from `git init`. This is creates the config file and finds the top-level directory for GN. Like `e fetch` in that it defaults to $PWD/electron, but you can also specify another directory.

`e build` -- Now bootstraps if necessary before building, so `e bootstrap` is redundant and has been removed. `e build --gen` exists in case you ever need to force a new `gn gen` before building.

So the new flow is `e init` -> `e sync` -> `e build`. If you really want to YOLO it, `e init --boostrap` does all three.

`evm` is now an `e` subcommand:

| nvm                  | e                  |
|:---------------------|:-------------------|
| nvm ls               | e show configs     |
| nvm current          | e show current     |
| nvm use &lt;name&gt; | e use &lt;name&gt; |

Those are the functional changes. The rest of the subcommands are all ported from the bash scripts, e.g. Shelley's patch exporter and PR URL opener. :tada: 

## What the help looks like

```
# e --help
Usage: e <command> [commandArgs...]

Electron build tool

Options:
  -h, --help                   output usage information

Commands:
  init|new [options] <name>    Create a new build config
  sync [gclientArgs...]        Get / update source code
  build|make [options]         Build Electron and other things
  show <subcommand>            Show info about the current build config
  test [specRunnerArgs...]     Run Electron's spec runner
  pr                           Open a GitHub URL where you can PR your changes
  patches <basename>           Refresh the patches in $root/src/electron/patches/$basename
  start|run [electronArgs...]  Run the Electron build
  node [electronArgs...]       Run the Electron build as Node
  debug                        Run the Electron build in a debugger
  use <name>                   Use build config <name> when running other `e` commands
  help [cmd]                   display help for [cmd]
```

```$ e help init
Usage: e-init [options] <name>

Create a new build config

Options:
  -r, --root <path>    Source and build files will be stored in this new directory (default: "/home/charles/electron/build-tools/electron")
  -i, --import <name>  Import build settings from $root/src/electron/build/args/$import.gn (default: "testing")
  -o, --out <name>     Built files will be placed in $root/src/out/$out (default: "Testing")
  -f, --force          Overwrite existing build config with that name (default: false)
  --asan               When building, enable clang's address sanitizer (default: false)
  --tsan               When building, enable clang's thread sanitizer (default: false)
  --msan               When building, enable clang's memory sanitizer (default: false)
  --lsan               When building, enable clang's leak sanitizer (default: false)
  --bootstrap          Run `e sync` and `e build` after creating the build config.
  --use-https          During `e sync`, set remote origins with https://github... URLs instead of git@github... (default: false)
  -h, --help           output usage information
```


ckerr:
Finally requesting review now that (I think) Sam's back from holiday :)

src/e-patches.js
```diff
@@ -0,0 +1,39 @@
...
+      node: path.resolve(srcdir, 'third_party', 'electron_node'),
+      v8: path.resolve(srcdir, 'v8'),
+    };
+    if (!targets[target]) {
+      console.log(`${color.err} Unrecognized dir ${color.path(target)}.`);
+      console.log(`${color.err} Supported dirs: ${Object.keys(targets).join(', ')}`);
+      process.exit(1);
+    }
+    childProcess.execFileSync(
+      path.resolve(srcdir, 'electron', 'script', 'git-export-packages'),
```
codebytere: s/`git-export-packages`/`git-export-patches` :)

src/e-pr.js
```diff
@@ -0,0 +1,42 @@
...
+function guessPRTarget(config) {
+  const package = require(path.resolve(config.root, 'src', 'electron', 'package.json'));
+  if (package.version.includes('nightly')) {
+    return 'master';
+  }
+  const match = /^([0-9]+)\.([0-9]+)\.[0-9]+.*$/.exec(package.version);
+  if (match) {
+    return `${match[1]}-${match[2]}-x`;
+  }
+  throw `Failed to determine target PR branch`;
```
codebytere: Same as above - can we make this more actionable? This failing bc it didn't match the regex, so we should tell the users that

src/e-show.js
```diff
@@ -0,0 +1,160 @@
...
+
+program
+  .command('configs')
+  .alias('ls')
+  .description('Installed build config')
+  .action(() => {
+    let current;
+    try {
+      current = evmConfig.currentName();
+    } catch (e) {
```
codebytere: catch bindings are optional, so we can just do: `catch {}`

src/e-show.js
```diff
@@ -0,0 +1,160 @@
...
+      // maybe there is no current config
+    }
+    try {
+      const names = evmConfig.names();
+      if (names.length === 0) {
+        console.log('No build configs found.');
```
codebytere:
```suggestion
        console.log('No build configs found; have you created one?');
```

tools/zsh/_e
```diff
@@ -0,0 +1,101 @@
...
+  "$@" --help | \
+  awk '/^Commands:/{flag=1;next}/^$/{flag=0}flag' | \
+  sed 's/  /;/g' | cut -f2 -d';' | \
+  sed 's/|/\n/' | \
+  cut -f1 -d' ' | \
+  while read -r subcommand; do
+    subcommands+=($subcommand)
+  done
+  __e_subcommands=("$subcommands[@]")
+}
```
codebytere: i'd really like some comments for the code in this file, it's somewhat opaque

src/util.js
```diff
@@ -0,0 +1,169 @@
...
+  const opts = { env: config.env, stdio: 'ignore' };
+
+  if (os.platform() === 'win32') {
+    console.debug(`Building on Windows -- skipping ${color.path(sccache)}`);
+    return;
+  }
+
+  try {
+    childProcess.execFileSync(sccache, ['--stop-server'], opts);
+  } catch (e) {
```
codebytere: optional bindings here - `catch {}`

src/util.js
```diff
@@ -0,0 +1,169 @@
...
+    // it's OK for this to fail -- maybe it wasn't running
+  }
+
+  for (;;) {
```
codebytere: nit: personally tend to prefer `while (true)` for readability on intentional infinite loops

src/e
```diff
@@ -0,0 +1,72 @@
...
+      childProcess.execFileSync(exec, args, opts);
+    } catch (e) {
+      fatal(e);
+    }
+  });
+
+program
+  .command('node [electronArgs...]')
+  .allowUnknownOption()
+  .description('Run the Electron build as Node')
```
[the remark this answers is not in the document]
ckerr: Changed d7d0ea9a6213e897d375ef618b95a8e6ee7e42f1

src/util.js
ckerr: Is this a personal preference or idiomatic JS? If the latter I'm fine to make the change.

Generally I use `;;` because then we _know_ that the system isn't wasting cycles re-evaluating `true` each time we go through the loop

src/util.js
ckerr: see above

tools/zsh/_e
ckerr: These autocomplete scripts are _all_ flaming nightmares. I'll add some comments but it's still going to be somewhat opaque :smiley: 

Added 691fdeb4803b682a2a224742ed767c1d5877371d

src/e-show.js
ckerr: Changed 691fdeb4803b682a2a224742ed767c1d5877371d

src/e-pr.js
ckerr: Oops, I fixed this before seeing this comment, so it's addressed in a force-push. :roll_eyes:

src/e-patches.js
ckerr: Fixed 691fdeb4803b682a2a224742ed767c1d5877371d

src/e-init.js
```diff
@@ -0,0 +1,157 @@
...
+      throw `Build config ${color.config(name)} already exists! (${color.path(filename)})`;
+    }
+
+    // save it
+    const config = createConfig(name, options);
+    ensureRoot(config);
+    evmConfig.save(name, config);
+    console.log(`New build config ${color.config(name)} created in ${color.path(filename)}`);
+
+    // use it
```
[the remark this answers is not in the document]
ckerr: Changed 691fdeb4803b682a2a224742ed767c1d5877371d

src/e-init.js
```diff
@@ -0,0 +1,157 @@
...
+
+function init(name, options) {
+  try {
+    // make sure it's new
+    const filename = evmConfig.pathOf(name);
+    if (!options.force && fs.existsSync(filename)) {
+      throw `Build config ${color.config(name)} already exists! (${color.path(filename)})`;
+    }
+
+    // save it
```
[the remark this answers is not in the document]
ckerr: I've changed this passage to "save the new config" and "`e use` the new config."

Changed 691fdeb4803b682a2a224742ed767c1d5877371d

src/e-debug.js
```diff
@@ -0,0 +1,40 @@
...
+}
+
+try {
+  const config = evmConfig.current();
+  if (commandExistsSync('gdb')) {
+    run_gdb(config);
+  } else if (commandExistsSync('lldb')) {
+    run_lldb(config);
+  } else {
+    throw `No debugger found!`;
```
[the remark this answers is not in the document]
ckerr: Oops, I fixed this before seeing this comment, so it's addressed in a force-push. :roll_eyes:

src/e-build.js
```diff
@@ -0,0 +1,78 @@
...
+  const exec = os.platform === 'win32' ? 'ninja.exe' : 'ninja';
+  const args = [...ninjaArgs, target];
+  const opts = { cwd: evmConfig.outDir(config) };
+  depot.execFileSync(config, exec, args, opts);
+}
+
+program
+  .allowUnknownOption()
+  .arguments('[target] [ninjaArgs...]')
+  .description('Build Electron and other things.')
```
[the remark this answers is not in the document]
ckerr: "other executables" isn't perfect, since targets include things like "node:headers"

I've changed this to "Build Electron and other targets" as a cue to tie it to the other uses of 'target' in the help message, e.g.`--list-targets`

Changed d7d0ea9a6213e897d375ef618b95a8e6ee7e42f1

src/e
```diff
@@ -0,0 +1,72 @@
...
+        stdio: ['ignore', 'inherit', 'ignore'],
+      };
+      console.log(color.childExec(exec, args, opts));
+      childProcess.execFileSync(exec, args, opts);
+    } catch (e) {
+      fatal(e);
+    }
+  });
+
+program.command('debug', 'Run the Electron build in a debugger');
```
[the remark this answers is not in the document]
ckerr: Changed d7d0ea9a6213e897d375ef618b95a8e6ee7e42f1

src/e
```diff
@@ -0,0 +1,72 @@
...
+
+program
+  .command('init [options] <name>', 'Create a new build config')
+  .alias('new')
+  .command('sync [gclientArgs...]', 'Get / update source code')
+  .command('build [options]', 'Build Electron and other things')
+  .alias('make')
+  .command('show <subcommand>', 'Show info about the current build config')
+  .command('test [specRunnerArgs...]', `Run Electron's spec runner`)
+  .command('pr', 'Open a GitHub URL where you can PR your changes')
```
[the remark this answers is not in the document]
ckerr: Are you suggesting that we eliminate the guess-which-branch-this-is code that's already in build-tools, or that we allow it to be overridden?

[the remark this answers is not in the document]
ckerr: I changed it s.t. the guessed versions are the default, but that it can be overridden via the command line. eede4017b8a8f0a402b59c89a99c4d85b616860f

src/e-init.js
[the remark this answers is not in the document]
ckerr: Extra context: the other way to go, wrt making this more descriptive, is to describe what's being saved and where. However this is spelled out explicitly in the console.log() a few lines below, so I didn't want to throw in redundant information

src/util.js
ckerr: Ah no I'm wrong! I misunderstood what you were saying here and above. Fixed 807e7651ff453ebf7b1b8a5700b7e12d4ecbeee3

src/util.js
codebytere: ah fair! no, i just think it's slightly less readable but there's no meaningful need to change :D

PR #26, opened 2019-10-31
chore: add some example config files

codebytere:
Adds some example config files for reference.

cc @MarshallOfSound

### DIFF
--- a/example-configs/evm.base.yml
+++ b/example-configs/evm.base.yml
@@ -1,0 +1,17 @@
+root: /path/to/your/developer/folder
+origin:
+  # ssh: git@github.com:electron/electron.git
+  # https: https://github.com/electron/electron
+  electron: https_or_ssh_url
+  # ssh: git@github.com:electron/node.git
+  # https: https://github.com/nodejs/node
+  node: https_or_ssh_url
+gen:
+  args:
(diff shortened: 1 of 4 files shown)


example-configs/evm.base.yml
```diff
@@ -0,0 +1,18 @@
+# Th
```
MarshallOfSound: Th?

example-configs/evm.base.yml
```diff
@@ -0,0 +1,18 @@
...
+  node: https_or_ssh_url
+gen:
+  args:
+    # path to sccache for faster builds (https://github.com/mozilla/sccache)
+    - cc_wrapper="path/to/your/developer/folder/electron/src/electron/external_binaries/sccache"
+env:
+  GIT_CACHE_PATH: /Users/use_name/.git_cache
+  SCCACHE_BUCKET: electronjs-sccache-ci
+  SCCACHE_CACHE_SIZE: 40G
+  SCCACHE_TWO_TIER: true
```
MarshallOfSound: New line

example-configs/evm.base.yml
codebytere: derp, deleted

Issue #28, opened 2019-11-01
e start doesn't properly parse directory arguments


codebytere:
When running `e start my-electron-app` locally, only the default app runs.

When i log `program.rawArgs` it looks like:
```
repros ❯ e start my-electron-app                                      7:40PM
[
  '/Users/codebytere/.nvm/versions/node/v12.7.0/bin/node',
  '/Users/codebytere/build-tools/src/e',
  'start',
  'my-electron-app'
]
```

so i'm fairly certain in this case we'll want to grab or parse these options differently 🤔

PR #29, opened 2019-11-01
fix: properly pass extra cli options

codebytere:
Closes https://github.com/electron/build-tools/issues/28.

`.allowUnknownOptions()` allows unknown options to be passed to a command without throwing an error, but with one twist - it only considers options to be those that start with `--`. That means that if we were to run `e start my-electron-app` it would consider `my-electron-app` to be an `arg`, and so with current code would just never be passed to `execFileSync`, since we were assuming all non-specified strings after the command would get pushed into `unknown`.

Also added some help text to both `e node` and `e start`.

cc @MarshallOfSound and @ckerr

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -11,7 +11,7 @@ jobs:
       uses: actions/setup-node@v1
       with:
         node-version: 12.x
-    - name: Lint bash code with shellcheck
+    - name: Lint
       run: |
         npm install
         npm run lint
--- a/src/e
+++ b/src/e
@@ -11,35 +11,43 @@ program.description('Electron build tool').usage('<command> [commandArgs...]');
 program
   .command('init [options] <name>', 'Create a new build config')
   .alias('new')
-  .command('sync [gclientArgs...]', 'Get / update source code')
+  .command('sync [gclientArgs...]', 'Get or update source code')
   .command('build [options]', 'Build Electron and other things')
   .alias('make');
 
 program
-  .command('start [electronArgs...]')
+  .command('start')
   .alias('run')
+  .description('Run the Electron executable')
   .allowUnknownOption()
-  .description('Run the Electron build')
   .action(() => {
     try {
       const exec = evmConfig.execOf(evmConfig.current());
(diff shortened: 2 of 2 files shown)


src/e
```diff
@@ -4,42 +4,50 @@ const childProcess = require('child_process');
...
-  .command('node [electronArgs...]')
-  .allowUnknownOption()
+  .command('node')
   .description('Run the Electron build as if it were a Node.js executable')
+  .allowUnknownOption()
   .action(() => {
     try {
       const exec = evmConfig.execOf(evmConfig.current());
-      const args = program.parseOptions(program.rawArgs.slice(2)).unknown;
+      const args = getFilteredArgs(program.rawArgs.slice(2));
```
ckerr: Same question here

src/util.js
```diff
@@ -13,6 +13,12 @@ function resolvePath(p) {
   return path.resolve(process.cwd(), p);
 }
 
+function getFilteredArgs(rawArgs) {
```
ckerr: If this function gets merged (see other question) but it's only used in `e`, probably better to put it there until its needed by other scripts?

.github/workflows/lint.yml
```diff
@@ -11,7 +11,7 @@ jobs:
       uses: actions/setup-node@v1
       with:
         node-version: 12.x
-    - name: Lint bash code with shellcheck
+    - name: Lint
```
ckerr: Nice catch :smiley:

src/util.js
codebytere: o yes good point! will consolidate 🙇‍♀

src/e
```diff
@@ -4,42 +4,50 @@ const childProcess = require('child_process');
...
+  .command('start')
   .alias('run')
+  .description('Run the Electron executable')
   .allowUnknownOption()
-  .description('Run the Electron build')
   .action(() => {
     try {
       const exec = evmConfig.execOf(evmConfig.current());
-      const args = program.parseOptions(program.rawArgs.slice(2)).unknown;
+      const args = getFilteredArgs(program.rawArgs.slice(2));
```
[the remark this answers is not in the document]
codebytere: no haha you're right - changing 😁

PR #32, opened 2019-11-07
feat: add build time telemetry

loc:
This adds telemetry for elapsed build times. I mostly wanted this so that we can give expected full and incremental build times in the docs and inform a future "recommended dev machine specs" section (as discussed at the Electron Summit '19).

I chose an un-authed database, Firebase, because I didn't want to deal with anything more complicated. If someone abuses my account, no big deal, there are spending limits and the data isn't important. However, if we have anything like this already set up (I remember seeing a Grafana graph at one point?), then it'd be easy to switch out.

I anticipate being able to distinguish between incremental and full builds with some basic clustering.

The machine-specific UUID is included in case we want to collect more information about devs' machines later. I threw in what is easily available from Node's `os` to make rudimentary analysis a bit quicker.

Here is a sample record:
```
{
  build_target: "electron",
  cores: 12,
  elapsed_time_ms: 94448,
  electron_version: "7.1.0",
  ram: 34359738368,
  timestamp: 1573087941782,
  uid: "e83b335b-2c79-4d24-9f21-46393a165321"
}
```

Some people might not want this info recorded for some reason; they can `set EVM_SKIP_TELEMETRY=1` or `export EVM_SKIP_TELEMETRY=1`.

### DIFF
--- a/src/e-build.js
+++ b/src/e-build.js
@@ -5,9 +5,10 @@ const fs = require('fs');
 const os = require('os');
 const path = require('path');
 const program = require('commander');
+const https = require('https');
 
 const evmConfig = require('./evm-config');
-const { depot, sccache, fatal } = require('./util');
+const { depot, sccache, fatal, getOrCreateUuid, readElectronVersion } = require('./util');
 
 function runGNGen(config) {
   depot.ensure();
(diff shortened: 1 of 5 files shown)


src/e-build.js
```diff
@@ -33,7 +34,37 @@ function runNinja(config, target, ninjaArgs) {
...
+    build_target: target,
+    electron_version: readElectronVersion(config),
+  });
+}
+
+function recordBuildTiming({ elapsed_time_ms, build_target, electron_version }) {
+  if (process.env.EVM_SKIP_TELEMETRY) return;
+  // If it fails, nbd, we'll just miss this timing.
+  const req = https.request({
+    hostname: 'electron-build-perf-21c0b.firebaseio.com',
```
MarshallOfSound: For multiple reasons this should go through an endpoint in electron/data and get piped into our grafana instance

MarshallOfSound:
Closing this out as we're beyond the "conflicted" state.  Happy to give this another pass if a new PR arrives and further discussion is had

PR #34, opened 2019-11-07
feat: fail gracefully on unknown commands

codebytere:
Previously, unknown commands would fail silently; no help text would be output and nothing would happen - this might be confusing to users.

This PR checks commands against a list of valid commands and outputs a help text if an unknown command is run.

Before:
Screen Shot 2019-11-07 at 2 05 55 PM

After:
Screen Shot 2019-11-07 at 2 05 29 PM

cc @MarshallOfSound @ckerr

### DIFF
--- a/src/e
+++ b/src/e
@@ -6,7 +6,19 @@ const program = require('commander');
 const evmConfig = require('./evm-config');
 const { color, fatal } = require('./util');
 
-program.description('Electron build tool').usage('<command> [commandArgs...]');
+program
+  .description('Electron build tool')
+  .usage('<command> [commandArgs...]')
+  .action(() => {
+    const command = program.rawArgs.slice(2)[0];
+    const isValid = item => item._name === command || item._alias === command;
+    if (!program.commands.find(item => isValid(item))) {
+      console.log('');
+      console.error(`Unrecognized command ${color.cmd(command)}`);
+      console.log('');
+      program.help();
+    }
+  });
 
 program
   .command('init [options] <name>', 'Create a new build config')
--- a/README.md
+++ b/README.md
@@ -217,8 +217,8 @@ along to the runner.
 # run all tests
 e test
 
-# Run main process tests in CI mode
-e test --ci --runners=main
+# Run main process tests
+e test --runners=main
 ```
 
 ## Getting Information
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
   "devDependencies": {
     "jest": "^24.9.0",
     "nyc": "^14.1.1",
-    "prettier": "^1.18.2",
+    "prettier": "^1.19.1",
     "rimraf": "^3.0.0"
   }
 }


codebytere:
@ckerr ☝️i tested that locally  and i'm not sure that it produces the behavior that we want? this means that no matter whether it succeeds or fails, it will output the help menu to console:

```sh
build-tools on git:fail-gracefully ❯ e node ~/Developer/electron-gn/src/third_party/e
lectron_node/test/sequential/test-inspector
Running "/Users/codebytere/Developer/electron-gn/src/out/Testing/Electron.app/Contents/MacOS/Electron /Users/codebytere/Developer/electron-gn/src/third_party/electron_node/test/sequential/test-inspector"
[test] Testing /json
[err] Debugger listening on ws://127.0.0.1:64739/3d16a7ff-08bd-48f9-ab34-08b95da900f5
[err] For help, see: https://nodejs.org/en/docs/inspector
[err] 
[test] Testing /json/list
[test] Testing /json/version
[test] Testing /json/activate
[test] Testing /json/activate/boom
[test] Testing /json/badpath
[test] Connecting to a child Node process
[test] Testing /json/list
[test] Verifying debugger stops on start (--inspect-brk option)
[err] Debugger attached.
[err] 
[test] Setting a breakpoint and verifying it is hit
[out] A message 5
[out] 
[test] Verify we can read current application state
[test] Verify sending and receiving UTF8 characters
[out] טֶ字и
[out] 
[out] A
[out] 
[out] B
[out] 
[test] Verify node waits for the frontend to disconnect
[out] Outputted message #1
[out] 
[err] Waiting for the debugger to disconnect...
[err] 
Unrecognized command: node

Usage: e <command> [commandArgs...]

Electron build tool

Options:
  -h, --help                 output usage information

Commands:
  init|new [options] <name>  Create a new build config
  sync [gclientArgs...]      Get or update source code
  build|make [options]       Build Electron and other things
  start|run                  Run the Electron executable
  node                       Run the Electron build as if it were a Node.js executable
  debug                      Run the Electron build with a debugger (gdb or lldb)
  use <name>                 Use build config <name> when running other `e` commands
  show <subcommand>          Show info about the current build config
  test [specRunnerArgs...]   Run Electron's spec runner
  pr [options]               Open a GitHub URL where you can PR your changes
  patches <basename>         Refresh the patches in $root/src/electron/patches/$basename
  help [cmd]                 display help for [cmd]

See https://github.com/electron/build-tools/blob/master/README.md for usage.
```

ckerr:
Oh, ugh, you're right. It doesn't look like the return value of parse() is something we can rely on.

I see two possibilities: one, we go with something more like your patch. Two, we use the `commander.commands` array something like this:

```js
program.parse(process.argv);
const command = program.rawArgs.slice(2)[0];
if (!program.commands.find(item => item._name === command || item._alias === command)) {
  console.error(`Unrecognized command ${color.cmd(command)}`);
  program.help();
}
```

I don't really like reinventing the wheel, but this second option makes assumptions about commander's internals which could change in the future.

src/e
```diff
@@ -4,20 +4,34 @@ const childProcess = require('child_process');
...
 
-program.description('Electron build tool').usage('<command> [commandArgs...]');
+program
+  .description('Electron build tool')
+  .usage('<command> [commandArgs...]')
+  .action(() => {
+    const command = program.rawArgs.slice(2)[0];
+    if (!Object.values(commands).includes(command)) {
+      console.log('');
+      console.log(`Unrecognized command: ${command}.`);
```
ckerr:
```suggestion
      console.log(`Unrecognized command: ${colors.cmd(command)}.`);
```

src/e
```diff
@@ -4,20 +4,34 @@ const childProcess = require('child_process');
...
+      console.log('');
+      console.log(`Run:`);
+      console.log('  $ e --help');
+      console.log('');
+      console.log('To see usage information.');
```
ckerr: I would replace this block just with `program.help();`

dsanders11:
> Previously, unknown commands would fail silently; no help text would be output and nothing would happen - this might be confusing to users.

I agree with @codebytere, it's not great behavior to not have feedback on unknown commands, speaking as a user.

It looks like this functionality is baked into `commander` 5/6, so upgrading that dependency would get the desired result. I did a quick test and it does fix that issue, but doesn't pass tests due to some changes to argument/option handling. Looking at their changelog it doesn't seem like any of the breaking changes are very big, so would just need someone more familiar with how it's currently used to take a look and adjust as necessary.

PR #36, opened 2019-11-14
fix: e-patches crashes from missing config setting

ckerr:
Looks like the config object structure was refactored a little and this didn't get the memo. childProcess was also missing, which is concerning since it implies the previous commit couldn't've worked regardless of config.
    
It might be worthwhile to give e-patches.js a little love in a followup PR in the slow-tests section. Since it's unusable right now, I'm PR'ingthis separately.

CC @codebytere

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+*       @ckerr @MarshallOfSound @codebytere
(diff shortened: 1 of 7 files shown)


ckerr:
Well. I really should fetch more often.

PR #38, opened 2019-11-18
chore: use yarn everywhere

codebytere:
Previously, we committed our `yarn.lock` but used `npm` both in CI and in documentation, which created some issues because npm and yarn don't always resolve identical versions.

This ensure we use yarn everywhere.

cc @ckerr

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -13,5 +13,5 @@ jobs:
         node-version: 12.x
     - name: Lint
       run: |
-        npm install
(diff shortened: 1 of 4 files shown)


codebytere:
lol ok will change!

ckerr:
I'm fine standardizing on yarn but when doing so we should also document the prerequisite in README.md

codebytere:
@ckerr added a note - please let me know if you think it's lacking or you'd like to convey the prereq some other way!

PR #45, opened 2020-01-05
docs: clarify new directory needed

codebytere:
Make it clearer that `--root=/path/to/new/electron/directory ` is an optional override path.

cc @ckerr @MarshallOfSound

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ e load-macos-sdks
 # Choose the directory where Electron's source and build files will reside.
 # You can specify any path you like; this command defaults to ~/projects/electron.
 # If you're going to use multiple branches, you may want something like:
-# `--root=$PWD/electron/branch` (e.g. `~/electron-gn/master`)
+# `--root=~/electron/branch` (e.g. `~/electron-gn/master`)
 e init --root=$PWD/electron --bootstrap testing
 ```
 
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ $Env:Path += ";$(pwd)"
 
 ## Getting the Code and Building Electron
 
-After installing build-tools, you can run a new Electron build with this command:
+You can run a new Electron build with this command:
 
 ```sh
 # The 'Hello, World!' of build-tools: get and build `master`
@@ -49,7 +49,11 @@ After installing build-tools, you can run a new Electron build with this command
 # On macOS if you aren't sure you have the right SDK version you should run
 e load-macos-sdks
 
-e init --root=/path/to/new/electron/directory --bootstrap testing
+# Choose the directory where Electron's source and build files will reside.
+# You can specify any path you like; this command defaults to ~/projects/electron.
+# If you're going to use multiple branches, you may want something like:
+# `--root=$PWD/electron/branch` (e.g. `~/electron-gn/master`)
+e init --root=$PWD/electron --bootstrap testing
 ```
 
 That command's going to run for awhile. While you're waiting, grab a
@@ -116,7 +120,7 @@ As an example, let's say you're starting from scratch and want both
 testing and release builds of the master branch in `electron/electron`.
 You might do this:
 
-```
+```sh
 # making 'release' and 'testing' builds from master
 
 $ e init master-testing -i testing --root=~/src/electron
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ e load-macos-sdks
 # You can specify any path you like; this command defaults to ~/projects/electron.
 # If you're going to use multiple branches, you may want something like:
 # `--root=~/electron/branch` (e.g. `~/electron-gn/master`)
-e init --root=$PWD/electron --bootstrap testing
+e init --root=~/electron --bootstrap testing
 ```
 
 That command's going to run for awhile. While you're waiting, grab a


README.md
```diff
@@ -41,14 +41,32 @@ $Env:Path += ";$(pwd)"
...
+# get its path (on Unix):
+cd electron-gn && pwd
+
+# get its path (on Windows):
+cd electron-gn && cd
+```
+
+in the directory where you typically place your code projects.
+
+After you've created this new directory, you can run a new Electron build with this command:
```
ckerr: This part isn't necessary -- `e init` will try to create the directory if it doesn't already exist.

In the literal case of `/path/to/new/electron/directory` dying with an error -- that's happening because there's no permission to create `/path`. If the user had used `/tmp/path/to/new/electron/directory` instead, that would have succeeded.

README.md
```diff
@@ -41,14 +41,32 @@ $Env:Path += ";$(pwd)"
...
 
 ```sh
 # The 'Hello, World!' of build-tools: get and build `master`
 
 # On macOS if you aren't sure you have the right SDK version you should run
 e load-macos-sdks
 
+# REPLACE /path/to/new/electron/directory with the path to the directory
+# you created in the command you ran above
```
ckerr: :+1: on adding this clarification, but needs to be reworked a little if the previous section is removed (see comment above).

Maybe something like

```
# Choose where where to keep Electron's source and build files.
# You can choose any naming scheme you like. Two options are
# * `--init=~/electron-gn` if you're only going to use one branch
# * `--init=~/electron-gn/branch` (e.g. `~/electron-gn/master`) if you're going to use multiple branches
```

?

README.md
```diff
@@ -41,14 +41,18 @@ $Env:Path += ";$(pwd)"
...
 # The 'Hello, World!' of build-tools: get and build `master`
 
 # On macOS if you aren't sure you have the right SDK version you should run
 e load-macos-sdks
 
+# Choose the directory where Electron's source and build files will reside.
+# You can specify any path you like. Two possibilities are
+# * `--root=~/electron-gn` if you're only going to use one branch
+# * `--root=~/electron-gn/branch` (e.g. `~/electron-gn/master`) if you're going to use multiple branches
 e init --root=/path/to/new/electron/directory --bootstrap testing
```
MarshallOfSound: Is there a reason we can't have this copy-pasteable as `e init --root=~/projects/electron ...` and let folks customize it if they want instead of forcing folks to basically type this out

ckerr: Not a bad idea

codebytere: done

README.md
```diff
@@ -41,15 +41,19 @@ $Env:Path += ";$(pwd)"
...
 
 ```sh
 # The 'Hello, World!' of build-tools: get and build `master`
 
 # On macOS if you aren't sure you have the right SDK version you should run
 e load-macos-sdks
 
-e init --root=/path/to/new/electron/directory --bootstrap testing
+# Choose the directory where Electron's source and build files will reside.
+# You can specify any path you like; this command defaults to ~/projects/electron.
```
ckerr: This is incorrect; it defaults to `$PWD/electron`

README.md
```diff
@@ -41,15 +41,19 @@ $Env:Path += ";$(pwd)"
...
+# If you're going to use multiple branches, you may want something like:
+# * `--root=~/electron-gn/branch` (e.g. `~/electron-gn/master`)
+e init --root=~/projects/electron --bootstrap testing
```
ckerr: I'd suggest we use consistent directories here instead of mixing `~/electron-gn` and `~/projects/electron` as the base.  Maybe something like

```text
# If you're going to use multiple branches, you may want something like
# `--root=~/electron-gn/branch` (e.g. `~/electron-gn/master`)
e init --root=~/electron-gn --bootstrap testing
```

README.md
```diff
@@ -41,15 +41,19 @@ $Env:Path += ";$(pwd)"
...
 # The 'Hello, World!' of build-tools: get and build `master`
 
 # On macOS if you aren't sure you have the right SDK version you should run
 e load-macos-sdks
 
-e init --root=/path/to/new/electron/directory --bootstrap testing
+# Choose the directory where Electron's source and build files will reside.
+# You can specify any path you like; this command defaults to ~/projects/electron.
+# If you're going to use multiple branches, you may want something like:
+# * `--root=~/electron-gn/branch` (e.g. `~/electron-gn/master`)
```
ckerr: The leading `*` is a copy paste error?

bnb:
+1, just ran into this 👀

README.md
```diff
@@ -41,15 +41,19 @@ $Env:Path += ";$(pwd)"
...
 # The 'Hello, World!' of build-tools: get and build `master`
 
 # On macOS if you aren't sure you have the right SDK version you should run
 e load-macos-sdks
 
-e init --root=/path/to/new/electron/directory --bootstrap testing
+# Choose the directory where Electron's source and build files will reside.
+# You can specify any path you like; this command defaults to ~/projects/electron.
+# If you're going to use multiple branches, you may want something like:
+# `--root=$PWD/electron/branch` (e.g. `~/electron-gn/master`)
```
ckerr:
```suggestion
# `--root=~/electron/branch` (e.g. `~/electron-gn/master`)
```

README.md
```diff
@@ -41,15 +41,19 @@ $Env:Path += ";$(pwd)"
...
+e init --root=$PWD/electron --bootstrap testing
```
ckerr:
```suggestion
e init --root=~/electron --bootstrap testing
```

PR #46, opened 2020-01-10
fix: correctly guess target branch for >= 8

codebytere:
Fixes `e pr` for branches following the new `8-x-y` pattern.

Previously, this command would try to open against `8-0-x` for a pr against `8-x-y`, which is incorrect.

cc @ckerr @MarshallOfSound

### DIFF
--- a/src/e-pr.js
+++ b/src/e-pr.js
@@ -8,19 +8,24 @@ const open = require('open');
 const program = require('commander');
 
 const evmConfig = require('./evm-config');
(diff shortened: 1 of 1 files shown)


src/e-pr.js
```diff
@@ -8,19 +8,24 @@ const open = require('open');
 const program = require('commander');
 
 const evmConfig = require('./evm-config');
-const { fatal } = require('./util');
```
codebytere: unused import

PR #47, opened 2020-01-14
feat: add configuration options for Goma

codebytere:
This PR adds configuration options for use with the Goma distibuted build system. This approach assumes that we'll be performing the authentication step for Goma at init-time, and then if we've stated we are using goma but are not authed properly an error is thrown at build-time. 

cc @MarshallOfSound @jkleinsc

### DIFF
--- a/src/e-build.js
+++ b/src/e-build.js
@@ -7,7 +7,7 @@ const path = require('path');
 const program = require('commander');
 
 const evmConfig = require('./evm-config');
-const { depot, sccache, fatal } = require('./util');
+const { depot, sccache, fatal, goma } = require('./util');
 
 function runGNGen(config) {
   depot.ensure();
@@ -26,7 +26,15 @@ function ensureGNGen(config) {
 }
 
 function runNinja(config, target, ninjaArgs) {
-  sccache.ensure(config);
+  if (config.goma) {
+    const authenticated = goma.isAuthenticated(config.root);
+    if (!authenticated) {
+      throw new Error('Goma not authenticated.');
+    }
+  } else {
+    sccache.ensure(config);
+  }
+
   depot.ensure(config);
   ensureGNGen(config);
 
--- a/src/e-build.js
+++ b/src/e-build.js
@@ -27,9 +27,13 @@ function ensureGNGen(config) {
 
 function runNinja(config, target, ninjaArgs) {
   if (config.goma) {
-    const authenticated = goma.isAuthenticated(config.root);
-    if (!authenticated) {
-      throw new Error('Goma not authenticated.');
+    if (goma.exists(config.root)) {
+      const authenticated = goma.isAuthenticated(config.root);
+      if (!authenticated) {
+        throw new Error('Goma not authenticated.');
+      } else {
+        goma.ensure(config.root);
+      }
     }
   } else {
     sccache.ensure(config);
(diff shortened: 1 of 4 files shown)


src/util.js
```diff
@@ -178,6 +197,11 @@ module.exports = {
   ensureDir,
   fatal,
   resolvePath,
+  authenticateGoma,
```
MarshallOfSound: We don't need this right?

src/e-build.js
```diff
@@ -26,7 +26,15 @@ function ensureGNGen(config) {
 }
 
 function runNinja(config, target, ninjaArgs) {
-  sccache.ensure(config);
+  if (config.goma) {
+    const authenticated = goma.isAuthenticated(config.root);
+    if (!authenticated) {
+      throw new Error('Goma not authenticated.');
+    }
+  } else {
```
MarshallOfSound: We need a call here to `goma.ensure` that runs `goma_ctl.py ensure_start`

src/e-init.js
```diff
@@ -16,15 +16,19 @@ function getVSReleaseLine() {
...
 
   // build the `gn gen` args
-  const gn_args = [
-    `import("//electron/build/args/${options.import}.gn")`,
-    `cc_wrapper="${sccache.exec(root)}"`,
-  ];
+  const gn_args = [`import("//electron/build/args/${options.import}.gn")`];
+
+  if (options.useGoma) {
+    gn_args.push('import("//electron/build/args/goma.gn")');
```
MarshallOfSound: We should only do this if this file actually exists.  `build-tools` still needs to work for < master

src/util.js
```diff
@@ -105,8 +104,46 @@ function depotExecFileSync(config, exec, args, opts_in) {
   childProcess.execFileSync(exec, args, opts);
 }
 
+const getExternalBinaries = root => path.resolve(root, 'src', 'electron', 'external_binaries');
+
+function gomaIsAuthenticated(root) {
+  const gomaDir = path.resolve(getExternalBinaries(root), 'goma');
+
+  // Bail early if we're not on a branch with the ability to use Goma
+  if (fs.existsSync(gomaDir)) return;
```
MarshallOfSound:
```suggestion
  if (!fs.existsSync(gomaDir)) return;
```

src/util.js
```diff
@@ -105,8 +104,46 @@ function depotExecFileSync(config, exec, args, opts_in) {
...
+
+  const loggedInPattern = /^Login as (\w+\s\w+)$/;
+  return loggedInPattern.test(loggedInInfo.toString().trim());
+}
+
+function authenticateGoma(root) {
+  const gomaDir = path.resolve(getExternalBinaries(root), 'goma');
+
+  // Bail early if we're not on a branch with the ability to use Goma
+  if (fs.existsSync(gomaDir)) return;
```
MarshallOfSound:
```suggestion
  if (!fs.existsSync(gomaDir)) return;
```

src/util.js
```diff
@@ -105,8 +104,46 @@ function depotExecFileSync(config, exec, args, opts_in) {
...
+  if (!gomaIsAuthenticated(root)) {
+    childProcess.execFileSync('python', ['goma_auth.py', 'login'], { cwd: gomaDir });
+  }
+}
+
+function ensureGomaStart(root) {
+  const gomaDir = path.resolve(getExternalBinaries(root), 'goma');
+
+  // Bail early if we're not on a branch with the ability to use Goma
+  if (fs.existsSync(gomaDir)) return;
```
MarshallOfSound:
```suggestion
  if (!fs.existsSync(gomaDir)) return;
```

codebytere:
hammering a few things out, hold for a few :)

PR #48, opened 2020-01-14
feat: add support for downloading a fixed and known version of xcode

MarshallOfSound:
This is so that Goma cache hit rates will be higher when using build tools.  The xcode version being used is a DEFINE in the electron build so we need to all be on the same version to get the best cache rates.  Currently this is using 10.3


MarshallOfSound:
@ckerr Do you mean running this during `init`?

PR #53, opened 2020-01-22
feat: add unknown command handler to run depot_tools cmds

MarshallOfSound:
Sometimes you want to be able to run `depot_tools` like `gn` manually, this lets you do it.

It's as easy as

```bash
e gn gen
# or
e goma_ctl status
```

### DIFF
--- a/src/e
+++ b/src/e
@@ -5,6 +5,8 @@ const program = require('commander');
 
 const evmConfig = require('./evm-config');
 const { color, fatal } = require('./utils/logging');
+const depot = require('./utils/depot-tools');
+const goma = require('./utils/goma');
 
 program.description('Electron build tool').usage('<command> [commandArgs...]');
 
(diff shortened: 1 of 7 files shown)


ckerr:
Specifically for the case of `e gn gen`, this is already handled with `e build --gen`

ckerr:
@MarshallOfSound since there seems to be some consensus on making this an explicit command, could you re-add this as an `e-depot.js` (or some other name) instead of having this as a fallback?

src/e
```diff
@@ -90,6 +92,32 @@ program
...
 
+program
+  .command('depot-tools')
+  .alias('d')
+  .description('Run a command from the depot-tools directory with the correct configuration')
+  .allowUnknownOption()
+  .action(() => {
+    const args = process.argv.slice(3);
+    let cwd;
+    if (args[0] === 'goma_ctl' || args[1] === 'goma_auth') {
```
ckerr: * should test to confirm args.length > 0

* more importantly, I _think_ you meant that `args[1]` to be `args[0]`?

src/e
```diff
@@ -90,6 +92,32 @@ program
...
+      cwd = goma.dir(evmConfig.current().root);
+      args[0] = `${args[0]}.py`;
+      args.unshift('python');
+    }
+
+    const { status, error } = depot.spawnSync(evmConfig.current(), args[0], args.slice(1), {
+      cwd,
```
ckerr: (Is it safe to pass an undefined `cwd` to spawnSyncs' options?)

MarshallOfSound: Yes 👍

PR #58, opened 2020-01-22
feat: let users of `e pr` specify a target repo

ckerr:
This lets users specify another repository to use. The default comes from the build-tools config file's `origin.electron`, i.e. `electron/electron`.
    
**IMPORTANT:** While this PR is probably worthwhile on its own, it is NOT a full fix for the issues raised at the training session:

  1. the user has to know to use the `--repo` argument
  2. it doesn't add a new config setting for the repo
  2. it doesn't address url rewriting at all

TBH this here mostly because I mentioned at the training session that I had this change in progress and didn't want anyone to think they needed to block on it. I'm going to be working on my presentation tonight and won't be around to work on build-tools :neutral_face:

### DIFF
--- a/src/e
+++ b/src/e
@@ -5,6 +5,8 @@ const program = require('commander');
 
 const evmConfig = require('./evm-config');
 const { color, fatal } = require('./utils/logging');
+const depot = require('./utils/depot-tools');
+const goma = require('./utils/goma');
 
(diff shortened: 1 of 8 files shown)


ckerr:
Whoaaa apparently my branch is not in sync with master. This was supposed to change one file, not eight. Closing this PR.

PR #62, opened 2020-01-23
fix: quick argument sanitization for 'e d'

MarshallOfSound:
cc @ckerr


codebytere:
Wait - lint died

ckerr:
*shakes fist at lint*

PR #66, opened 2020-01-28
feat: default to goma enabled, add support for cache-only goma

MarshallOfSound:
This does a few things:

* Enables goma by default
* Goma now has two modes "cluster" and "cache-only"
* Completely removes sccache as the cache will not be populated any more
* Makes build-tools responsible for downloading and controlling goma.  This way we can do what Chromium does and upgrade / change goma versions outside of version control for electron itself

I tested this on my machine with:

```bash
e init --root=/Users/sattard/projects/electron/electron --bootstrap rolling --goma=cluster
```

Which completed in 18 minutes 😮 

Fixes: #67.


groundwater:
`s/goma/not goma/`

src/util.js
```diff
@@ -0,0 +1,150 @@
...
+const macOSSDKsPath = path.resolve(__dirname, '..', 'third_party', 'macOS_SDKs');
+const DEPOT_TOOLS_DIR = process.env.DEPOT_TOOLS_DIR || defaultDepotPath;
+
+function resolvePath(p) {
+  if (path.isAbsolute(p)) return p;
+  if (p.startsWith('~/')) return path.resolve(os.homedir(), p.substr(2));
+  return path.resolve(process.cwd(), p);
+}
+
+function ensureDir(dir) {
```
codebytere: this file looks like it was accidentally added?

src/e-init.js
```diff
@@ -145,10 +129,15 @@ program
...
-    `Use Electron's custom deployment of Goma (only available to maintainers).`,
+    '--disable-goma',
+    'Disables usage of Goma, this will make the build take substantially longer',
     false,
   )
+  .option(
+    '--goma <target>',
+    `Use Electron's custom deployment of Goma.  Can be "cache-only" or "cluster".  The "cluster" mode is only available to maintainers`,
+    'cache-only',
+  )
```
ckerr: These two options seem to overlap.

What would you think of having three targets: `cluster`, `cache-only`, or `none`, and removing the `--disable-goma` option?

src/e-init.js
```diff
@@ -173,6 +162,17 @@ try {
...
   }
 
+  // Make sure the goma options are valid
+  if (options.disableGoma) {
+    options.goma = false;
+  } else if (!['cache-only', 'cluster'].includes(options.goma)) {
+    throw new Error(
+      `Config property ${color.config('goma')} must be one of ${color.path(
+        'cache-only',
+      )} or ${color.path('cluster')} but you provided ${color.path(options.goma)}`,
```
ckerr: _nit_ these three `color.path` above should probably be `color.config`

src/e-init.js
```diff
@@ -186,11 +186,13 @@ try {
 
   // (maybe) run sync to ensure external binaries are downloaded
   if (program.bootstrap) {
-    childProcess.execFileSync('node', [e, 'sync', '-v'], opts);
+    childProcess.execFileSync('node', [e, 'sync', '-v', '--ignore_locks'], opts);
```
ckerr: :heart:

src/e-show.js
```diff
@@ -152,13 +152,12 @@ program
 
 program
   .command('stats')
-  .description('sccache build statistics')
+  .description('goma build statistics')
```
ckerr:
```suggestion
  .description('build statistics')
```

src/utils/goma.js
```diff
@@ -1,16 +1,92 @@
...
+        sha,
+      )}. Halting now`,
+    );
+    rimraf.sync(tmpDownload);
+    process.exit(1);
+  }
+
+  if (filename.endsWith('.tgz')) {
+    const result = childProcess.spawnSync('tar', ['zxvf', filename], {
+      cwd: path.resolve(tmpDownload, '..'),
```
ckerr: _nit_

we use `path.resolve(tmpDownload, '..')` in both the `if` and `else` clauses... maybe move it to the line before `if`

```js
const extractPath = path.dirname(tmpDownload);
if (...
```

MarshallOfSound:
Added some helpers for folks with older configs

image

PR #73, opened 2020-01-31
feat: add support for custom ninja target

MarshallOfSound:
For building other projects or just more specific custom targets in Electron.

This also adds support for a new `defaultTarget` config so that other projects can specify what target gets built when you just run `e build`

### DIFF
--- a/src/e-build.js
+++ b/src/e-build.js
@@ -6,7 +6,7 @@ const path = require('path');
 const program = require('commander');
 
 const evmConfig = require('./evm-config');
-const { fatal } = require('./utils/logging');
+const { color, fatal } = require('./utils/logging');
 const depot = require('./utils/depot-tools');
 const goma = require('./utils/goma');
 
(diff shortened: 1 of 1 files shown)


src/e-build.js
```diff
@@ -64,38 +64,41 @@ program
...
   const config = evmConfig.current();
+
+  const pretty_targets = {
+    breakpad: 'third_party/breakpad:dump_sym',
+    chromedriver: 'electron:electron_chromedriver_zip',
+    electron: 'electron',
+    'electron:dist': 'electron:electron_dist_zip',
+    mksnapshot: 'electron:electron_mksnapshot_zip',
+    'node:headers': 'third_party/electron_node:headers',
+    __default: config.defaultTarget || 'electron',
```
ckerr: This double-underscore prefix has me confused -- is this a user-facing key that one would type in on the command line (e.g. `e build __default`), or is it an internal placeholder for when one types `e build`?

If it's intended as a user-facing key, could we use `default` instead? The leading double underscore makes this look like an implementation variable rather than a user-facing option and feels weird.

If it's intended as an internal placeholder, dumping this internal key in the console.log on line 86 is a leaky impl. We could bypass this by removing the `__default` lines altogether and doing this:

```js
const pretty = Object.keys(pretty_targets).find(p => program.rawArgs.includes(p));
...
const target = pretty_targets[pretty] || config.defaultTarget || 'electron';
runNinja(config, target, args);
```

Either of these approaches (public key named `default` or removing `*default` altogether) would be fine; I don't have a preference

MarshallOfSound: It's now a public "default" and I also improved the behavior of `--list-targets`

PR #75, opened 2020-01-31
feat: add support for EVM_SESSION_ID to have per-shell config selection

### LINKED ISSUE #72: Support using two different configs in different terminals
I want to be able to `evm use foo` in one shell and `evm use bar` in the other shell and both to still work.  `nvm` supports this so technically it must be possible to set this kind of thing up

MarshallOfSound:
Fixes #72 

By adding `export EVM_SESSION_ID=$(uuid)` to your `.bashrc` or `.zshrc` file you can get per-session evm configs, so you can use different configs in different shells.

`e use X` will set both your session and the global config to `X`
`e <anything>` will attempt to read your session, if it fails it will read global and then instantly set that result to your session

### DIFF
--- a/README.md
+++ b/README.md
@@ -298,3 +298,16 @@ Valid patch directories include:
 [nvm]: https://github.com/nvm-sh/nvm
 [platform-prerequisites]: https://electronjs.org/docs/development/build-instructions-gn#platform-prerequisites
 [sanitizers]: https://github.com/google/sanitizers
+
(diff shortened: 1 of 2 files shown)


MarshallOfSound:
Superseded by #76

ckerr:
...I opened #76 as a PR to this, rather than master -- I wasn't trying to supersede this PR

PR #76, opened 2020-02-01
refactor: use EVM_CURRENT_FILE to avoid fs clutter

ckerr:
The intent is to follow the same workflow as nvm: e has a default active config (just as nvm has a default active version of Node.js), but that can be overridden by the user on a per-shell basis.

What's the same as EVM_SESSION_ID: since `e` is a JS script and can't modify its caller's environment variables, the goal is accomplished by having more than one 'current' file. The default file is $EVM_CONFIG/evm-current.txt. An optional environment variable specifies where to find an optional file to override the default.

What's different from EVM_SESSION_ID: EVM_CURRENT_FILE specifies the filename itself, so evm-config.js doesn't need new folder code and $EVM_CONFIG/sessions/ doesn't accumulate dead session folders. Instead, the file can be in $TMPDIR so that the OS will clean it up later. Also, the `EVM_CURRENT` code has also been removed since this new feature does the same thing in a way that's more like nvm.
    
To opt in to this behavior, one would add something like `export EVM_CURRENT_FILE="$(mktemp --tmpdir evm-current.XXXXXXXXX.txt)"` to one's ~/\.[profile|bashrc|zshrc].

### DIFF
--- a/src/evm-config.js
+++ b/src/evm-config.js
@@ -7,12 +7,16 @@ const { color } = require('./utils/logging');
 const { ensureDir } = require('./utils/paths');
 const goma = require('./utils/goma');
 
+const preferredFormat = process.env.EVM_FORMAT || 'json'; // yaml yml json
 const configRoot = process.env.EVM_CONFIG || path.resolve(__dirname, '..', 'configs');
+
+// If you want your shell sessions to each have different active configs,
+// try this in your ~/.profile or ~/.zshrc or ~/.bashrc:
+// export EVM_CURRENT_FILE="$(mktemp --tmpdir evm-current.XXXXXXXX.txt)"
 const currentFiles = _.compact([
   process.env.EVM_CURRENT_FILE,
   path.resolve(configRoot, 'evm-current.txt'),
 ]);
-const preferredFormat = process.env.EVM_FORMAT || 'json'; // yaml yml json
 
 function buildPath(name, suffix) {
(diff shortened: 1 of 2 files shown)


src/evm-config.js
```diff
@@ -7,13 +7,16 @@ const { color } = require('./utils/logging');
...
-let sessionFile;
-if (process.env.EVM_SESSION_ID && !process.env.EVM_SESSION_ID.includes('.')) {
-  sessionFile = path.resolve(configRoot, 'sessions', process.env.EVM_SESSION_ID, 'evm-current.txt');
-}
 const preferredFormat = process.env.EVM_FORMAT || 'json'; // yaml yml json
+const configRoot = process.env.EVM_CONFIG || path.resolve(__dirname, '..', 'configs');
+
+// If you want your shell sessions to each have different active configs,
+// try this in your ~/.profile or ~/.zshrc or ~/.bashrc:
+// export EVM_CURRENT_FILE="$(mktemp --tmpdir evm-current.XXXXXXXX.txt)"
```
codebytere: I would prefer we document this instead of leaving it in a comment - otherwise someone would only be able to find the functionality by looking at this specific file

ckerr: I held off on doing that because it seemed like too much information for a README, but you're right that it needs to be documented.

In 9a46e841785b1188a2b5fc533b0615e8a4e48d59 I've added an 'Advanced Usage' section and made EVM_CURRENT_FILE a subitem underneath that

PR #77, opened 2020-02-03
docs: improve documentation verbosity

codebytere:
Add a little more information about some of the command-specific documention and clean up examples. 

cc @ckerr @MarshallOfSound

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,16 +164,31 @@ If you're starting from scratch, this will (slowly) fetch all the source
 code. It's also useful after switching Electron branches to synchronize
 the rest of the sources to the versions needed by the new Electron branch.
 
-`e sync` is usually all you need. Any extra args are passed along to gclient,
-so for example `e sync -v` runs gclient verbosely.
+`e sync` is usually all you need. Any extra args are passed along to gclient itself.
 
 ```sh
 $ e show current
 master-testing
+
 $ e show root
 ~/src/electron
+
+$ e sync
+Running "gclient sync --with_branch_heads --with_tags" in '~/src/electron/src'
+[sync output omitted]
+```
+
+To make your output more verbose, you can add an increasing number of `-v`s. For example,
+
+```sh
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -258,36 +278,52 @@ e test --runners=main
...
 
-Valid patch directories include:
+Valid patch directories can include:
 
 * `node`
 * `v8`
 * `boringssl`
 * `chromium`
+* `perfetto`
+* `icu`
```
ckerr: Are these two correct? I don't see them in https://github.com/electron/electron/blob/master/patches/config.json

src/e-build.js
```diff
@@ -24,11 +24,11 @@ function runGNGen(config) {
...
-  const argsFile = path.resolve(evmConfig.outDir(config), 'args.gn');
-  if (!fs.existsSync(argsFile)) return runGNGen(config);
-  const contents = fs.readFileSync(argsFile, 'utf8');
-  // If the current args do not match the args file, re-run gen
-  if (contents.trim() !== config.gen.args.join('\n').trim()) return runGNGen(config);
+  // const argsFile = path.resolve(evmConfig.outDir(config), 'args.gn');
+  // if (!fs.existsSync(argsFile)) return runGNGen(config);
+  // const contents = fs.readFileSync(argsFile, 'utf8');
+  // // If the current args do not match the args file, re-run gen
+  // if (contents.trim() !== config.gen.args.join('\n').trim()) return runGNGen(config);
```
ckerr: Is this part intentional for a docs PR?

README.md
codebytere: @ckerr i included them as they exist on some older branches (see: https://github.com/electron/electron/pull/21969); our patches can run a range but i'd rather have them here for verbosity's sake so folks know they're a possibility

src/e-build.js
codebytere: ope

no

PR #97, opened 2020-02-19
fix: only load newer macOS sdks

codebytere:
We should only be linking 10.14 & 10.15, as older ones now cause compilation errors.

cc @MarshallOfSound

### DIFF
--- a/src/e-load-xcode.js
+++ b/src/e-load-xcode.js
@@ -16,7 +16,8 @@ if (process.platform !== 'darwin') {
 Xcode.ensureXcode();
 macOSSDKs.ensure();
 
-const SDK_TO_LINK = ['10.13', '10.14', '10.15'];
+const SDK_TO_LINK = ['10.14', '10.15'];
+const SDK_TO_UNLINK = ['10.12', '10.13'];
 
 const xCodeSDKDir = path.resolve(
   Xcode.XcodePath,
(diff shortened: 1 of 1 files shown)


MarshallOfSound:
Can we add a thing to unlink 10.13

src/e-load-xcode.js
```diff
@@ -48,6 +51,23 @@ for (const sdk of SDK_TO_LINK) {
   childProcess.execFileSync('sudo', ['ln', '-s', sourceSDK, targetDirectory]);
 }
 
+// Unlink unnecessary macOS SDKs.
+for (const sdk of SDK_TO_UNLINK) {
+  const sourceSDK = path.resolve(macOSSDKs.path, `MacOSX${sdk}.sdk`);
+  if (!fs.existsSync(sourceSDK)) continue;
+
+  const targetDirectory = path.resolve(xCodeSDKDir, `MacOSX${sdk}.sdk`);
+  if (!fs.existsSync(targetDirectory)) continue;
```
MarshallOfSound: Can we check that the target is actually a symbolic link, some version of `fs.statSync()` should work

src/e-load-xcode.js
```diff
@@ -48,6 +51,23 @@ for (const sdk of SDK_TO_LINK) {
...
+
+  console.warn(
+    `${color.info} Removing symbolic link from ${color.path(sourceSDK)} --> ${color.path(
+      targetDirectory,
+    )}`,
+  );
+
+  childProcess.execFileSync('sudo', ['unlink', sourceSDK]);
```
MarshallOfSound: I've just realized I don't think we need `sudo` for this any more but I'll check in a separate PR

MarshallOfSound: Yeah we don't need `sudo` for this https://github.com/electron/build-tools/pull/112/files

src/e-load-xcode.js
```diff
@@ -48,6 +53,29 @@ for (const sdk of SDK_TO_LINK) {
...
+  const stats = fs.lstatSync(targetDirectory);
+  if (!stats.isSymbolicLink()) return;
+
+  console.warn(
+    `${color.info} Removing symbolic link from ${color.path(sourceSDK)} --> ${color.path(
+      targetDirectory,
+    )}`,
+  );
+
+  childProcess.execFileSync('unlink', [sourceSDK]);
```
MarshallOfSound: Shouldn't this be `unlink', [targetDirectory])` ?

codebytere: derp, ye. fixed.

Issue #91, opened 2020-02-15
Goma hangs on first run


samuelmaddock:
OS: Windows 10 64-bit

If Goma hasn't been started yet, `e build` seems to hang when ensuring it's started.
```
D:\work\electron>e build
Running "goma_ctl.py ensure_start" in D:\work\build-tools\third_party\goma
Terminate batch job (Y/N)? y

D:\work\electron>e build
Running "goma_ctl.py ensure_start" in D:\work\build-tools\third_party\goma
Not logged in
Running "D:\work\build-tools\third_party\depot_tools\gn.bat gen "out/Testing" --args="import(\"//electron/build/args/testing.gn\") import(\"D:\work\build-tools\third_party\goma.gn\")"" in D:\work\electron\master\src
Done. Made 12190 targets from 2156 files in 12242ms
```

To reproduce, kill `compiler_proxy.exe` in the task manager then run `e build` again.

MarshallOfSound:
I've seen this on windows too, cc @jkleinsc who I think has a script somewhere that can correctly launch goma on windows without the hang

jkleinsc:
@marshallofsound https://github.com/electron/electron/blob/master/script/start-goma.ps1 is what we use for CI

PR #107, opened 2020-02-24
feat: make build-tools error-correcting on Windows

### LINKED ISSUE #102: Automatically fix global git config issues
We should detect if we're running in an interactive environment (stdin must be a valid pipe and `process.env.CI` should not be set) and if so prompt the user if they want us to automatically fix the issues with the git config that we detect.

### LINKED ISSUE #104: Successful build on first-run on clean Windows machine
Running `e init` on a clean Windows machine should work first time and self-correct missing dependencies along the way (by prompting the user or by auto running commands) so that it will build Electron with no prompts and with maximum goma cache hits

MarshallOfSound:
With these changes I can install Node, Git and `build-tools` and run 

`e init --root=~/electron --bootstrap testing` and it Just Works out of the box, it will prompt to install missing dependencies, fix incorrect git configuration, etc.

Closes #102 
Closes #104 

Currently stacked on top of #96 for Better Windows Reasons

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,8 @@
     "open": "^6.4.0",
     "path-key": "^3.1.0",
     "readline-sync": "^1.4.10",
-    "stream-progressbar": "^1.3.0"
+    "stream-progressbar": "^1.3.0",
+    "which": "^2.0.2"
   },
   "devDependencies": {
     "husky": "^4.2.0",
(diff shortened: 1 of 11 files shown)


MarshallOfSound:
The only thing that stops this being an end-to-end clean run is #91 which I'd like to address in a follow up

PR #136, opened 2020-04-07
feat: inherit current config

ckerr:
Change the recommendation on when to set `EVM_CURRENT_FILE` in the startup bashrc/zshrc scripts: only set `EVM_CURRENT_FILE` if it's not already set.

Most of the time this change won't have any affect.

However it makes life nicer if you're hacking on multiple Electron versions at the same time and want new shells that you create (e.g. new tabs in a terminal) to inherit the right version. By inheriting the `EVM_CURRENT_FILE` environment variable, new shells will be synced to their parent's version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -340,7 +340,9 @@ Valid patch directories can include:
 If you want your shell sessions to each have different active configs, try this in your `~/.profile` or `~/.zshrc` or `~/.bashrc`:
 
 ```sh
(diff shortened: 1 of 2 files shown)


MarshallOfSound:
If this impacts new tabs in terminal wouldn't this mean that a second tab can never have a different session to the first tab.  Having some kind of `EVM_PARENT_FILE` or something and defaulting to to that while still using a unique file for the new session might match expected behavior more?

PR #151, opened 2020-07-10
fix: don't compare number to string

codebytere:
String coercion,,,,,,how does it work,,,,,,,

cc @MarshallOfSound

### DIFF
--- a/src/e-pr.js
+++ b/src/e-pr.js
@@ -23,7 +23,7 @@ function guessPRTarget(config) {
     const [major, minor] = [match[1], match[2]];
 
     //TODO(codebytere): remove this conditional when 7-1-x is EOL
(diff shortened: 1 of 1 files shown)


src/e-pr.js
```diff
@@ -23,7 +23,7 @@ function guessPRTarget(config) {
     const [major, minor] = [match[1], match[2]];
 
     //TODO(codebytere): remove this conditional when 7-1-x is EOL
-    return major >= '8' ? `${major}-x-y` : `${major}-${minor}-x`;
+    return major >= 8 ? `${major}-x-y` : `${major}-${minor}-x`;
```
MarshallOfSound, line 26:
```suggestion
    return parseInt(major, 10) >= 8 ? `${major}-x-y` : `${major}-${minor}-x`;
```

PR #153, opened 2020-07-10
fix: re-initializing the same directory should just sync

### LINKED ISSUE #150: when running `e init` on a bootstrap that already exists, the path of how to progress is unclear
I've picked up build_tools a few times, either from fresh or after a few months. Each time, my brain goes to the same bootstrap name which ends up failing since it already exists. I can't tell if it being in some destroyed/incomplete state matters here, but that is usually the state of things when I'm trying to re-run it.

An example of the command I run once:
```
$ e init --root=~/electron --bootstrap local     
```

Then I get this: 
```
ERROR Error: Build config local already exists! (/Users/cyren/GitHub/electron/build-tools/configs/evm.local.json)
    at Object.<anonymous> (/Users/cyren/GitHub/electron/build-tools/src/e-init.js:152:11)
    at Module._compile (internal/modules/cjs/loader.js:1200:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1220:10)
    at Module.load (internal/modules/cjs/loader.js:1049:32)
    at Function.Module._load (internal/modules/cjs/loader.js:937:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:71:12)
    at internal/main/run_main_module.js:17:47
```

It would be nice if there was:
- a way to either pick up where I left off with that `e init`, or
- a way to wipe that config

MarshallOfSound:
Fixes #150 

If we run `init` with a config name that already exists, just let it fall through to sync if the config that already exists also points at the currently targeted "root" directory.

### DIFF
--- a/src/e-init.js
+++ b/src/e-init.js
@@ -145,10 +145,19 @@ try {
 
   checkPlatformDependencies();
 
+  const config = createConfig(options);
+
   // make sure the config name is new
   const filename = evmConfig.pathOf(name);
   if (!options.force && fs.existsSync(filename)) {
-    throw Error(`Build config ${color.config(name)} already exists! (${color.path(filename)})`);
+    const existing = evmConfig.fetchByName(name);
+    if (existing.root !== config.root) {
+      throw Error(
+        `Build config ${color.config(
+          name,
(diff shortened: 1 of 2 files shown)


bnb:
danke <3

PR #154, opened 2020-07-14
feat: add "e patches all" option

ckerr:
Real straightfoward, what this patch does is add an `e patches all` command which invokes `src/electron/script/export_all_patches.py` from the right directory and with the right config file.

### DIFF
--- a/src/e-patches.js
+++ b/src/e-patches.js
@@ -20,10 +20,23 @@ function exportPatches(target) {
       targets[path.basename(key)] = val;
     }
 
-    if (!targets[target]) {
+    if (target === 'all') {
+      const script = path.resolve(srcdir, 'electron', 'script', 'export_all_patches.py');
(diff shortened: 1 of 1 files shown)


ckerr:
CC @jkleinsc for inspiration :slightly_smiling_face:

src/e-patches.js
```diff
@@ -20,29 +20,37 @@ function exportPatches(target) {
...
   } catch (e) {
     fatal(e);
   }
 }
 
 program
-  .arguments('<basename>')
-  .description('Refresh the patches in $root/src/electron/patches/$basename')
+  .arguments('<basename-or-all>')
+  .description('Refresh all patches,  or the patches in $root/src/electron/patches/$basename')
```
codebytere, line 54: for clarity here, i think i'd rename the argument to `'<target>'` and then in the description say: `Refresh all patches if 'all' is specified, otherwise refresh patches in $root/src/electron/patches/$target'`

PR #160, opened 2020-08-24
fix: do not kill the build on cache-only mode if the server dies

MarshallOfSound:
I hope this fixes an issue where the function would randomly return a not-authed error instead of a 200 cache-miss response.  Until we figure out why the not-authed responses arrive this should fix the issue by making goma not self-kill on a 401.

### DIFF
--- a/src/e-build.js
+++ b/src/e-build.js
@@ -60,7 +60,7 @@ function runNinja(config, target, ninjaArgs) {
       }
     }
 
-    goma.ensure();
+    goma.ensure(config);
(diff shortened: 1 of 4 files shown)


codebytere:
```
    expect(received).toStrictEqual(expected) // deep equality

    Expected: 0
    Received: 1

      30 | 
      31 |       // confirm that it worked
    > 32 |       expect(result.exitCode).toStrictEqual(0);
         |                               ^
      33 |       expect(fs.statSync(root).isDirectory()).toStrictEqual(true);
      34 |       expect(fs.statSync(gclient_file).isFile()).toStrictEqual(true);
      35 |     });
```

PR #161, opened 2020-08-24
feat: add -v and --version

bnb:
Adds a simple version option. Pulls in the package.json and spits out the version property.


bnb:
Since build-tools is self-updating, it seems that the version isn't actually maintained and this is not a particularly useful addition - going to close 😅

PR #172, opened 2020-09-05
feat: improve Electron fork support

codebytere:
Improve support for working on Electron without write access using `build-tools`.

Previously, one's remotes could get overwritten during sync, and it wasn't clear how to push to a fork in the context of Electron within `src`. Did you need to clone the fork somehow? Something else? Fairly confusing.

This makes it such that a new user can now run e.g.:

```sh
$ e init --fork codebytere/electron --bootstrap testing
```

And then when inside `/src/electron`:
```sh
$ git remote -v
# origin  https://github.com/electron/electron (fetch)
# origin  https://github.com/electron/electron (push)
# fork    https://github.com/codebytere/electron (fetch)
# fork    https://github.com/codebytere/electron (push)
```

cc @MarshallOfSound @ckerr


MarshallOfSound (commented):
This will break existing configs / stop them working.  Can we support new configs and add a warning that folks should update their config.

MarshallOfSound (approved):

PR #183, opened 2020-09-22
feat: add remove config command

### LINKED ISSUE #182: Request: Add command for removing build config
Add convince command for cleaning up old build config files in `~/.electron_build_tools/configs`.

e.g. `e remove|rm <name>` (Prevent removal of the current build config).

holmberd:
This PR adds a new command for removing old build configurations.

- `e remove|rm <name>`
- Prevents the removal of the current used build config.

resolves #182

### DIFF
--- a/src/evm-config.js
+++ b/src/evm-config.js
@@ -203,6 +203,8 @@ function sanitizeConfig(config) {
 }
 
 function remove(name) {
+  testConfigExists(name);
+
   let currentConfigName;
   try {
     currentConfigName = currentName();
@@ -213,7 +215,6 @@ function remove(name) {
     throw Error(`Config is currently in use`);
   }
 
-  testConfigExists(name);
   const filename = pathOf(name);
   try {
     return fs.unlinkSync(filename);
--- a/src/evm-config.js
+++ b/src/evm-config.js
@@ -39,6 +39,16 @@ function filenameToConfigName(filename) {
   return match ? match[1] : null;
 }
 
+function ensureConfigExists(name) {
+  if (!fs.existsSync(pathOf(name))) {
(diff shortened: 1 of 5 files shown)


src/evm-config.js
```diff
@@ -198,6 +202,26 @@ function sanitizeConfig(config) {
   return config;
 }
 
+function remove(name) {
+  let currentConfigName;
+  try {
+    currentConfigName = currentName();
+  } catch (_) {
```
codebytere, line 209: nit: catch bindings are optional, we can just say 
```
catch {
  currentConfigName = null;
}
``` 
here

src/evm-config.js
```diff
@@ -39,6 +39,16 @@ function filenameToConfigName(filename) {
   return match ? match[1] : null;
 }
 
+function ensureConfigExists(name) {
```
ckerr, line 42: _(nit)_ Could this be renamed to something else, e.g. `testConfigExists()` I usually see 'ensureThing()' used  as "if thing is false, then make thing true" -- so to me this function name reads "if name doesn't exist, then create it"

codebytere:
@holmberd this is awesome, thanks! 🥳 

Could you please try to add a test or two? Once that's done we can get this merged.

ckerr (commented):
overall LGTM. I had a minor naming nit, and would like to see this added to the README.md docs as well, but otherwise looks good!

src/evm-config.js
holmberd: Yep that makes sense.

codebytere (approved):

ckerr (approved):
Thanks!

PR #187, opened 2020-09-23
fix: if sanitizeConfig() makes changes, save them.

ckerr:
Minor annoyance fix: whenever sanitizeConfig() makes changes,  save those changes back to disk so that you don't keep getting the warnings until you manually edit (or remove) the offending config. :smile_cat: 

CC @codebytere @MarshallOfSound

### DIFF
--- a/src/evm-config.js
+++ b/src/evm-config.js
@@ -198,14 +198,30 @@ function sanitizeConfig(config) {
   return config;
 }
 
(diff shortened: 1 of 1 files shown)


ckerr:
Just saw https://github.com/electron/build-tools/pull/186; IMO that's a better approach

PR #188, opened 2020-09-23
feat: add a hint on how to make changes permanent

ckerr:
When warning the user that changes were made, also give them a hint on how to make changes permanent.

Sample output of this change:

```
WARN We've made these temporary changes to your configuration:
 * replaced deprecated "origin" property with "remotes" property
See "e sanitize-config" to make these changes permanent.
```

CC @codebytere

### DIFF
--- a/src/evm-config.js
+++ b/src/evm-config.js
@@ -179,7 +179,7 @@ function sanitizeConfig(name, overwrite = false) {
     changes.push(`added ${color.config('CHROMIUM_BUILDTOOLS_PATH')} definition`);
   }
 
-  if (changes.length !== 0) {
+  if (changes.length > 0) {
     if (overwrite) {
       save(name, config);
     } else {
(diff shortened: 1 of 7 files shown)


src/evm-config.js
```diff
@@ -189,33 +170,23 @@ function sanitizeConfig(name, overwrite = false) {
...
-          'CHROMIUM_BUILDTOOLS_PATH',
-        )} - we've temporarily added it for you`,
-      );
-    }
+    changes.push(`added ${color.config('CHROMIUM_BUILDTOOLS_PATH')} definition`);
   }
 
-  if (overwrite) {
-    save(name, config);
+  if (changes.length !== 0) {
```
codebytere, line 182: (nit) for clarity's sake this should probably be `changes.length > 0`

codebytere (approved):
approved w nit!

src/evm-config.js
ckerr: changed d70c657

PR #197, opened 2020-10-10
feat: Added when packaging, auto remove dist.zip and strip electron.

### LINKED ISSUE #191: e build electron:dist doesn't remove symbol dist.zip is very large.
When I use
```
e build electron:dist -v
```
It generates dist.zip, but the size is greater than 1g. 
Can I add strip binaries by e command?

```
electron/script/strip-binaries.py -d out/Release
```

bigben0123:
First, if dist.zip exists, it cannot be replace when generating it again.
Second, when packaging dist.zip, the file electron is not stripped.
I fixed both of them. 
Fixes #191


MarshallOfSound (changes requested):
Sorry that this has just kind of sat here, I think this kind of change / build system fix should land into Electrons `BUILD.gn` file directly rather than being hacked around in build-tools

ckerr:
Closing this PR based on the comment above. If this is still an issue in `electron/electron`, maybe make a `BUILD.gn` PR there as suggested by Sam

PR #199, opened 2020-10-10
fix: resolve hang when starting goma

### LINKED ISSUE #91: Goma hangs on first run

dsanders11:
Fixes #91.

Now, don't ask me *why* it fixes it. 😄 Something to do with it hanging when stdio is actually attached. Didn't feel like spending more time debugging the root root cause once it was working.

@MarshallOfSound, I also have a branch that uses `start-goma.ps1` like you suggested, but that solution causes an extra window to appear briefly due to powershell, and that didn't seem ideal. Internally that script is also just using `goma_ctl.py ensure_start` so the extra layer of indirection didn't feel necessary one I resolved the hang.

### DIFF
--- a/src/utils/goma.js
+++ b/src/utils/goma.js
@@ -46,7 +46,7 @@ function downloadAndPrepareGoma(config) {
   if (fs.existsSync(path.resolve(gomaDir, 'goma_ctl.py'))) {
     depot.spawnSync(config, 'python', ['goma_ctl.py', 'stop'], {
       cwd: gomaDir,
-      stdio: 'ignore',
+      stdio: ['ignore'],
     });
   }
 
@@ -152,6 +152,7 @@ function ensureGomaStart(config) {
       ...process.env,
       ...gomaEnv(config),
     },
+    stdio: ['ignore'],
   });
 }
 


codebytere (commented):
Does this cause any behavioral changes on non-Windows devices - e.g should we do this only on Windows?

(also THANK YOU this has been driving me slowly nuts for ages)

dsanders11:
> Does this cause any behavioral changes on non-Windows devices - e.g should we do this only on Windows?

Not that I can think of, reasonably. Since `stdin` and `stdout` aren't being used for command, seems safe to ignore them all together. That said, I haven't tested it on a non-Windows platform as of yet, but I would be pretty surprised if there was a functional difference.

> (also THANK YOU this has been driving me slowly nuts for ages)

It only took coming back a second time thinking I'd started a build to find that it never began before I decided it had to be fixed at all costs, so I went nuts much faster. :smile:

codebytere (dismissed):

MarshallOfSound (commented):
Is it all stdio or can we just ignore stdin? Having the output of this command can be useful

dsanders11:
@MarshallOfSound, just tested, looks like `stdio: ['ignore', 'pipe', 'pipe']` also works. So yes, only ignoring stdin seems to fix it as well. Would you prefer that?

MarshallOfSound:
@dsanders11 Yup, just ignoring `stdin` is more optimal 👍

dsanders11:
@MarshallOfSound, updated the PR to do that. Looks like the invocation of `ensure_stop` was already using `ignore` and I hadn't noticed it before... that would have saved a little time. I updated it to also only ignore `stdin` for symmetry.

codebytere (approved):

codebytere:
@dsanders11 depending on your interest, feel free to fill out https://electronjs.org/maintainers/join for multi-channel guest access to our slack if you'd like to chat to us more!

PR #205, opened 2020-10-20
feat: --json flag for `show env` command

dsanders11:
For easy programmatic access to the environment.

### DIFF
--- a/src/e-show.js
+++ b/src/e-show.js
@@ -85,11 +85,18 @@ program
 program
   .command('env')
   .description('Environment variables set when building Electron')
-  .action(() => {
+  .option('--json', 'Output as JSON')
+  .action(options => {
     try {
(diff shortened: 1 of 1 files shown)


src/e-show.js
```diff
@@ -85,11 +85,18 @@ program
...
   .command('env')
   .description('Environment variables set when building Electron')
-  .action(() => {
+  .option('--json', 'Output as JSON')
+  .action(options => {
     try {
-      const exportKeyword = os.platform() === 'win32' ? 'set' : 'export';
-      const logger = ([key, val]) => console.log(`${exportKeyword} ${key}=${val}`);
-      Object.entries(evmConfig.current().env).forEach(logger);
+      const env = evmConfig.current().env;
```
codebytere, line 91: (nit) we can just destruct `const { env } - evmConfig.current()`

codebytere (approved):

PR #207, opened 2020-10-21
chore: run CI on all platforms

dsanders11:
Run CI on all the platforms... except MacOS for tests.

Couldn't get the MacOS tests to pass on CI, if anyone else wants to take a look.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -5,9 +5,23 @@ on: pull_request
 jobs:
   build:
     name: Test pull requests
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+      fail-fast: false
+    runs-on: ${{ matrix.os }}
     steps:
-    - uses: actions/checkout@master
+    - name: Set git global config (Windows)
+      if: ${{ matrix.os == 'windows-latest' }}
+      run: |
+        git config --global core.autocrlf false
+        git config --global core.eol lf
+        git config --global core.filemode false
+        git config --global branch.autosetuprebase always
+    - name: Install dependencies (Windows)
+      if: ${{ matrix.os == 'windows-latest' }}
+      run: pip install pywin32
+    - uses: actions/checkout@v1
       with:
         fetch-depth: 1
     - name: Use Node.js v12.x
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -3,9 +3,18 @@ name: Lint
 on: [push]
 
 jobs:
-  build:
-    runs-on: ubuntu-latest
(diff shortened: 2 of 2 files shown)


codebytere:
lol https://github.com/electron/build-tools/pull/206#issuecomment-713221204

.github/workflows/pull_request.yml
```diff
@@ -5,9 +5,28 @@ on: pull_request
 jobs:
   build:
     name: Test pull requests
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
```
codebytere, line 10: think this is missing `macos-latest`?

.github/workflows/pull_request.yml
```diff
@@ -5,9 +5,28 @@ on: pull_request
...
+    - name: Set git global config (Windows)
+      if: ${{ matrix.os == 'windows-latest' }}
+      run: |
+        git config --global core.autocrlf false
+        git config --global core.eol lf
+        git config --global core.filemode false
+        git config --global branch.autosetuprebase always
+    - name: Install dependencies (Windows)
+      if: ${{ matrix.os == 'windows-latest' }}
+      run: |
```
codebytere, line 23: since this is one line we can just do `run: pip install pywin32`

same below

codebytere:
Merged https://github.com/electron/build-tools/pull/206 - should be good to rebase

.github/workflows/pull_request.yml
dsanders11: Tests won't pass for some reason, it doesn't find files in the sandboxed directories it seems. So I left it off for now until someone can debug.

codebytere: weird - in that case should remove the macos-specific bits farther down then probably

dsanders11: Removed, and sure.

.github/workflows/pull_request.yml
dsanders11: Changed.

dsanders11:
@codebytere, rebased and good to go.

dsanders11:
Hmm, after the CI builds finished it was hung waiting on status to be reported. Found some info saying it might be related to `actions/checkout@v2`, which I had bumped to with this change. Changing back to `actions/checkout@v1` didn't fix it.

[This post](https://github.community/t/github-actions-pull-requests-always-include-some-outdated-checks/16157) suggests that possibly the 'Required' check is on the old non-matrix version of the action and so it won't ever return unless manually fixed?

codebytere:
@dsanders11 i fixed that up - it's a branch protection rule and github hasn't quite smoothed out their logic to detect when a required check no longer exists

PR #210, opened 2020-10-21
chore: ignore third_party js when formatting

codebytere:
Ignore `third_party` files when running `prettier:write`. Accidentally formatted all the minified files in `third_party/depot_tools/third_party/coverage/htmlfiles/` just now 😅 

cc @MarshallOfSound

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,1 @@
+third_party


MarshallOfSound (approved):

dsanders11:
@codebytere, did that happen on the git hook? Seems that the invocation in `scripts` wouldn't cover `third_party`. Seems like `lint-staged` in `package.json` should be using the script version so that it uses the correct paths?

codebytere:
this happens when you run `npm run prettier:write`

PR #211, opened 2020-10-21
feat: enable running specific Electron binaries

codebytere:
This makes it such that one can run specific electron binaries against a fork of electron-quick-start or similar repros which can't, for example, be written with Fiddle. I've been doing this myself for a while, and figured that it would be a handy thing to build into `build-tools`.

<details>
<Summary>New corresponding evm.base.yml</Summary>

```yaml
root: /Users/codebytere/Developer/electron-gn
electronBinaryDirectory: '/Users/codebytere/Library/Application Support/Electron Fiddle/electron-bin'
remotes:
  electron:
    origin: https://github.com/electron/electron
  node:
    origin: https://github.com/nodejs/node
env:
  GIT_CACHE_PATH: /Users/codebytere/.git_cache
  CHROMIUM_BUILDTOOLS_PATH: /Users/codebytere/Developer/electron-gn/src/buildtools
```

</details>

cc @ckerr

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,8 +247,10 @@ target create "/Users/yourname/electron-gn/src/out/Testing/Electron.app/Contents
 
 ### `e test`
 
-`e test ` starts the local Electron build's test runner. Any extra args are passed
+`e test ` starts the local Electron build's test runner by default. Any extra args are passed
 along to the runner.
+
+You can also optionally specify an Electron executable to run by passing the path to a directory containing an Electron executable, as well as set a configuration option allowing you to specify a directory containing multiple Electron executable files by setting `config.electronBinaryDirectory` and then passing `--v=x.y.z`.
 
 To see all potential options for this command, run:
 
@@ -265,7 +267,7 @@ e test
 e test --runners=main
 ```
 
-Possible extra arguments to pass:
+Possible extra arguments to pass to the runner:
 * `--node` - Run Node.js' own tests with Electron in `RUN_AS_NODE` mode.
 * `--runners=<main|remote|native>` - The set of tests to run, can be either `main`, `remote`, or `native`.
 
--- a/README.md
+++ b/README.md
@@ -247,10 +247,8 @@ target create "/Users/yourname/electron-gn/src/out/Testing/Electron.app/Contents
 
 ### `e test`
 
-`e test ` starts the local Electron build's test runner by default. Any extra args are passed
+`e test ` starts the local Electron build's test runner. Any extra args are passed
 along to the runner.
-
-You can also optionally specify an Electron executable to run by passing the path to a directory containing an Electron executable, as well as set a configuration option allowing you to specify a directory containing multiple Electron executable files by setting `config.electronBinaryDirectory` and then passing `--v=x.y.z`.
 
 To see all potential options for this command, run:
 
@@ -270,6 +268,21 @@ e test --runners=main
 Possible extra arguments to pass to the runner:
 * `--node` - Run Node.js' own tests with Electron in `RUN_AS_NODE` mode.
 * `--runners=<main|remote|native>` - The set of tests to run, can be either `main`, `remote`, or `native`.
+
+### `e bin`
+
+`e bin` allows for starting an app with a specific Electron executable.
+
+You can specify the Electron executable to run by passing the path to a directory containing an Electron executable, or by setting a configuration option allowing you to specify a directory containing multiple Electron executable files by setting `config.electronBinaryDirectory` and then passing `--v=x.y.z`.
+
+Example:
+```sh
+# start an app with Electron version 1.2.3
+e bin --v=1.2.3 /path/to/app
+
+# start an app with the Electron executable at a specific path
+e bin --path=/path/to/electron/exec /path/to/app
+```
 
 ### `e show`
 
--- a/src/e-start.js
+++ b/src/e-start.js
@@ -1,0 +1,77 @@
+#!/usr/bin/env node
+
+const childProcess = require('child_process');
+const path = require('path');
+const fs = require('fs');
+const semver = require('semver');
+const program = require('commander');
+
+const evmConfig = require('./evm-config');
+const { color, fatal } = require('./utils/logging');
+
+program
+  .allowUnknownOption()
+  .option(
+    '--v <version>',
+    'A specific version binary of Electron to run the specified app with. Must have set the binary directory in the current config.',
+  )
+  .option('--path <path>', 'Path to an Electron executable')
+  .on('--help', () => {
+    console.log('');
+    console.log('Examples:');
+    console.log('');
+    console.log('  $ e start .');
+    console.log('  $ e start /path/to/app');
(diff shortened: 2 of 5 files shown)


src/evm-config.js
```diff
@@ -150,8 +150,6 @@ function sanitizeConfig(name, overwrite = false) {
   }
 
   if (config.origin) {
-    const oldConfig = color.config(util.inspect({ origin: config.origin }));
```
codebytere, line 153: unused

src/e-start.js
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    args = program.rawArgs.slice(4);
+    exec = path.join(binaryPath, evmConfig.electronExecPath());
+  } else if (program.path) {
+    if (!fs.existsSync(program.path)) {
+      console.error(`${color.err} ${program.path} does not contain a valid Electron executable.`);
+      process.exit(1);
+    }
+    exec = path.join(program.path, evmConfig.electronExecPath());
```
ckerr, line 60: Seems like we should be building `args` here as well. This looks like the use case in the docs of `e start /path/to/exec --js-flags`

src/e-start.js
```diff
@@ -0,0 +1,71 @@
...
+  .parse(process.argv);
+
+try {
+  const config = evmConfig.current();
+
+  let exec;
+  let args;
+  if (program.v) {
+    const binDir = config.electronBinaryDirectory;
+    if (!binDir || !fs.existsSync(binDir)) {
```
ckerr, line 37:
```suggestion
    if (!binDir) {
```

This is one of those little things that adds up -- `!fs.existsSync(binDir)` gives false security here b/c it confirms there's something in that path, but doesn't confirm it's a directory -- so we don't actually know that binDir is valid when doing this validity test.

We could (a) leave it as a halfway test or (b) add try-catch-statSync(), which is sturdier but less concise, or (c) do nothing and let the `path.resolve()` test below handle it. IMO (c) is preferable since `path.resolve()` really is the true arbiter, so duplication is unnecessary.

Again, not a huge deal -- the code as-is would work. I've just learned to distrust half-tests as a red flag.

src/evm-config.js
```diff
@@ -162,7 +164,7 @@ function sanitizeConfig(name, overwrite = false) {
     };
 
     delete config.origin;
-    changes.push(`replaced superceded 'origin' property with 'remotes' property`);
+    changes.push(`replaced superseded 'origin' property with 'remotes' property`);
```
ckerr, line 167: ***sigh***

According to my commit logs I've been making this error for literally 20 years now. :older_adult: Apparently it's a blind spot.

ckerr (commented):
I **love** this PR. I've been doing this with `Electron Fiddle/electron-bin/` downloads by hand for awhile now but this is even better.

src/e-start.js
```diff
@@ -0,0 +1,71 @@
...
+      console.error(`${color.err} electronBinaryDirectory not defined in config or not valid`);
```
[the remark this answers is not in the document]
codebytere: We probably don't want to do this since we need to tell than exact _what_ is not defined - if `electronBinaryDirectory` this will log `undefined` which doesn't tell them what to fix

src/e-start.js
codebytere: oops, spaced that 😛

README.md
```diff
@@ -247,9 +247,11 @@ target create "/Users/yourname/electron-gn/src/out/Testing/Electron.app/Contents
 
 ### `e test`
 
-`e test ` starts the local Electron build's test runner. Any extra args are passed
+`e test ` starts the local Electron build's test runner by default. Any extra args are passed
 along to the runner.
 
+You can also optionally specify an Electron executable to run by passing the path to a directory containing an Electron executable, as well as set a configuration option allowing you to specify a directory containing multiple Electron executable files by setting `config.electronBinaryDirectory` and then passing `--v=x.y.z`.
```
dsanders11, line 253: Presumably this means manually modifying the config under ~/.electron_build_tools/configs/ to include that setting? That's not quite clear since there's no other mention of manually modifying configs (or where they are) in README.md and no commands for modifying configs.

codebytere: Yeah i thought about this a bit and tbh am not sure - if you were reading this, what do you think would make this clear to you? Or - what dx would you want to see?

src/e-start.js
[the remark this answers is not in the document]
ckerr: Oo you're right. I meant to type `'electronBinaryDirectory'` rather than `electronBinaryDirectory`

README.md
dsanders11: I think there should either be a way to modify the configs built into `e` (obviously that's a bit of work outside the scope of this), or a section of README.md which mentions where configs are stored for modifying them manually. Then this blurb could say: "by setting `config.electronBinaryDirectory` (see \[Modifying Configs](#modifying-configs))" or something like that.

Just my two cents. In the current form that's only going to be clear to people who have already poked around a bunch in `build-tools` since `config.electronBinaryDirectory` is pretty opaque without that knowledge.

ckerr: Not a blocker for this PR, but what I had in mind was something like `e start --fiddle=vX.Y.Z` and build-tools would Just Work by finding Fiddle's appData path automatically. This is my preference because then the script will do the work I'd be doing by hand on win/mac/linux.

This PR's mechanism is more universal, e.g. it doesn't require people to be using Fiddle in order to work, and I like that. IMO `--fiddle` could work by reusing some of this PR's code and would be very convenient for the use case of fiddle + buld-tools.

Adding `--fiddle` would require us to clone Electron's per-platform appData logic -- or more likely, to import it e.g from Sindre's env-paths package. I'd still be :+1: on this either in this PR if you're so inclined, or for any of us to write it in a followup PR.

MarshallOfSound (changes requested):
I'm not sure this should live on `e start`.  Having `e start` absorb certain arguments is gonna be a problem.  E.g. `e start --help`, does that show build tools help, or the `electron --help` screen?

This should probably be it's own command or something with a much more clearly defined _way of doing things_.  `e start` was originally designed as a literal proxy for `electron {...args}` changing that will mess with some things in un-workaroundable ways

ckerr:
> E.g. e start --help, does that show build tools help, or the electron --help screen?

The standard way of handling that use case is treating a standalone ` -- ` as the delimiter between the wrapper's args and the target's args; e.g. `e start --help` would show build-tools' help, but `e start -- --help` would show Electron's help.

I'm not opposed to doing this somewhere outside of `e start` but am not sure I see the point of decoupling? The `--` delimiter idiom seems to make sense here too.

MarshallOfSound:
> The -- delimiter idiom seems to make sense here too.

That `--` delimiter is just going to make the default 99% case of "running the electron binary I just built" harder.  Instead of `e start -i` I have to do `e start -- -i` which just seems wrong.  I think slicing this into `e fiddle` or `e bin` or something else will make sense.

codebytere:
I'd be fine with e bin i think - let me rework this a tad to that end.

PR #215, opened 2020-10-21
chore: only inject Windows env vars on Windows

dsanders11:
I don't believe there's any reason to have those on other platforms?

### DIFF
--- a/src/utils/depot-tools.js
+++ b/src/utils/depot-tools.js
@@ -41,6 +41,25 @@ function ensureDepotTools() {
   }
 }
 
+function platformOpts() {
+  let opts = {};
+
+  switch (os.platform()) {
+    case 'win32':
+      opts = {
+        DEPOT_TOOLS_WIN_TOOLCHAIN: '1',
+        DEPOT_TOOLS_WIN_TOOLCHAIN_BASE_URL:
+          'https://electron-build-tools.s3-us-west-2.amazonaws.com/win32/toolchains/_',
+        GYP_MSVS_HASH_9ff60e43ba91947baca460d0ca3b1b980c3a2c23:
+          '6d205e765a23d3cbe0fcc8d1191ae406d8bf9c04',
+        GYP_MSVS_HASH_a687d8e2e4114d9015eb550e1b156af21381faac:
+          'b1bdbc45421e4e0ff0584c4dbe583e93b046a411',
+      };
+  }
+
+  return opts;
+}
+
 function depotOpts(config, opts = {}) {
   // some defaults
   opts = {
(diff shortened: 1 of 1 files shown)


codebytere (approved):
agreed, ty

src/utils/depot-tools.js
```diff
@@ -54,13 +54,17 @@ function depotOpts(config, opts = {}) {
...
+      ? {
+          DEPOT_TOOLS_WIN_TOOLCHAIN: '1',
+          DEPOT_TOOLS_WIN_TOOLCHAIN_BASE_URL:
+            'https://electron-build-tools.s3-us-west-2.amazonaws.com/win32/toolchains/_',
+          GYP_MSVS_HASH_9ff60e43ba91947baca460d0ca3b1b980c3a2c23:
+            '6d205e765a23d3cbe0fcc8d1191ae406d8bf9c04',
+          GYP_MSVS_HASH_a687d8e2e4114d9015eb550e1b156af21381faac:
+            'b1bdbc45421e4e0ff0584c4dbe583e93b046a411',
+        }
+      : {}),
```
ckerr, line 67: :+1: on making this platform dependent. To bikeshed slightly though -- this PR takes an already-slightly-awkward code flow and makes it slightly _more_ awkward.

What would you think about something like this, extracting this out to another method for readability:

```js
  function platformOpts() {
    switch (os.platform()) {
      case 'win32':
        return {
          DEPOT_TOOLS_WIN_TOOLCHAIN: '1',
          DEPOT_TOOLS_WIN_TOOLCHAIN_BASE_URL:
            'https://electron-build-tools.s3-us-west-2.amazonaws.com/win32/toolchains/_',
          GYP_MSVS_HASH_9ff60e43ba91947baca460d0ca3b1b980c3a2c23:
            '6d205e765a23d3cbe0fcc8d1191ae406d8bf9c04',
          GYP_MSVS_HASH_a687d8e2e4114d9015eb550e1b156af21381faac:
            'b1bdbc45421e4e0ff0584c4dbe583e93b046a411',
        };
      default:
        return {};
    }   
  }
  
  function depotOpts(config, opts = {}) {
    ...
    opts.env = {
      // set these defaults that can be overridden via process.env
      PYTHONDONTWRITEBYTECODE: '1', // depot needs it
      DEPOT_TOOLS_METRICS: '0', // disable depot metrics
      ...process.env,
      ...platformOpts(),
      ...config.env,
      ...opts.env,
      // Circular reference so we have to delay load
      ...require('./goma').env(config),
    };
    ...
```

Not a blocker, but have a slight preference for this ^ for readability

dsanders11: That would be fine by me. I generally try to keep my PRs as surgical as possible to increase odds of getting a quick merge, so the current version was the least change to accomplish it. I agree that `depotOpts`'s code for `opts.env` is getting a bit awkward, so pulling it out seems logical. If everyone else is happy with that change I can make it. @codebytere

codebytere: +1 from me!

dsanders11:
@codebytere @ckerr, changes made. Was going to add test coverage but realized I'd need #216 to land first for where I was going to add it in `e-show-spec.js`.

codebytere (approved):

MarshallOfSound (approved):

PR #225, opened 2020-10-30
feat: add backport functionality to `e pr`

dsanders11:
This is built on top of #222. ~~I'll rebase once (if) that gets merged, so leaving as draft for now.~~

Borrows [some code from electron/clerk](https://github.com/electron/clerk/blob/20f4ef1cb723ba84a1a6f3c2e90dd9b51f1d2ee3/src/note-utils.ts#L7-L26). I didn't see an easy way to share that code so opted for copy-and-paste.

~~Not sure what the feasible upper-limit is for injecting the release notes in the URL. I tested with PR 25064 and it could handle that 12 word sentence. I think for practical purposes it's probably fine.~~

EDIT: Remembered that Vuetify uses the same functionality for their issues (you create it on a separate site and it redirects to the GitHub issue with it filled out). Just tested it with two paragraphs of 'lorem ipsum' and that worked fine, so the notes length isn't a concern.

### DIFF
--- a/src/e-pr.js
+++ b/src/e-pr.js
@@ -3,11 +3,55 @@
 const childProcess = require('child_process');
 const fs = require('fs');
 const path = require('path');
+const querystring = require('querystring');
 
+const got = require('got');
 const open = require('open');
 const program = require('commander');
 
 const evmConfig = require('./evm-config');
+const { color, fatal } = require('./utils/logging');
+
+// Adapted from https://github.com/electron/clerk
+function findNoteInPRBody(body) {
+  const onelineMatch = /(?:(?:\r?\n)|^)notes: (.+?)(?:(?:\r?\n)|$)/gi.exec(body);
+  const multilineMatch = /(?:(?:\r?\n)Notes:(?:\r?\n+)((?:\*.+(?:(?:\r?\n)|$))+))/gi.exec(body);
+
+  let notes = null;
+  if (onelineMatch && onelineMatch[1]) {
+    notes = onelineMatch[1];
+  } else if (multilineMatch && multilineMatch[1]) {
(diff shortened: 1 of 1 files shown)


dsanders11:
Rebased and ready for review.

dsanders11:
Rebased again.

src/e-pr.js
```diff
@@ -3,11 +3,55 @@
...
+
+  const opts = {
+    url: `https://api.github.com/repos/electron/electron/pulls/${pullNumber}`,
+    responseType: 'json',
+    throwHttpErrors: false,
+  };
+  try {
+    const response = await got(opts);
+    if (response.statusCode !== 200) {
+      fatal(`Could not find PR: ${opts.url} got ${response.headers.status}`);
```
codebytere, line 46: i'd rather `console.log` this with an error and not throw an error with a stack i think 🤔 

```js
console.log(color.err, Could not find PR: ${opts.url} got ${response.headers.status});
```

codebytere (approved):
smallish nit but otherwise lgtm!

src/e-pr.js
dsanders11: I think that's actually functionally what's happening with the current code. It looks like `fatal` will only include a stack if an error is provided, if a string is used (like here), then it will do what you suggested, just using `console.error` instead:

```javascript
function fatal(e) {
  if (typeof e === 'string') {
    console.error(`${color.err} ${e}`);
  } else {
    console.error(`${color.err} ${e.stack ? e.stack : e.message}`);
  }
  process.exit(1);
}
```

codebytere: ope wait you're totally right, i forgot we changed that!

PR #227, opened 2020-10-30
chore: run CI tests on push

dsanders11:
Only lint is running on push to `master` at the moment, running test as well might help catch a mix-up like #221 more quickly.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,10 +1,16 @@
-name: Continous Integration
+name: Continuous Integration
 
-on: pull_request
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -1,10 +1,10 @@
 name: Continous Integration
 
-on: pull_request
+on: [push, pull_request]
```
codebytere, line 3: We can be more specific; let's do:

```yml
on:
  push:
    branches:
      - master
  pull_request:
    branches:
      - master
```

codebytere (changes requested):

.github/workflows/test.yml
dsanders11: 👍

codebytere (approved):
ty!

Issue #229, opened 2020-11-02
Using Goma Can Slow Down Local Builds


dsanders11:
A downside to always going through Goma for building is that by default it limits the number of local subprocesses to a suboptimal number for local building. In situations where the Goma cluster isn't being used, such as no Internet access, or building something which will have a lot of cache misses, this means you're not using all available CPU power.

There's a handful of Goma environment variables which can be tweaked to get around this. I haven't really looked into which ones are actually useful for this end goal: `GOMA_EXCLUSIVE_NUM_PROCS`, `GOMA_MAX_SUBPROCS`, `GOMA_MAX_SUBPROCS_LOW`, `GOMA_MAX_SUBPROCS_HEAVY`,  `GOMA_BURST_MAX_SUBPROCS_HEAVY`, `GOMA_MAX_SUBPROCS_PENDING`.

[Here is the Goma code](https://chromium.googlesource.com/infra/goma/client/+/master/client/goma_flags.cc) which is pulling less than optimal numbers. You can see that for a 4-core CPU, the two functions will return 1 and 2 respectively. On a 4-core you end up limited to two processes for local building.

```c++
static int MaxSubProcsLow() {
  int cpus = devtools_goma::GetNumCPUs();
  if (cpus > 0)
    return std::max(1, cpus / 5);
  return 1;
}
static int MaxSubProcs() {
  // +1 is for high priority task like local fallback.
  return MaxSubProcsLow() + 1;
}
```

Since `-j` is set to 200 for Goma, and only affects how many Goma tasks will be used, a different flag (and default) for max number of local subprocesses would be good.

codebytere:
Addressed in https://github.com/electron/build-tools/pull/256

PR #230, opened 2020-11-02
feat: --no-goma flag for `e build`

dsanders11:
Refs #229.

Sometimes it's better to not use Goma, like when you don't have Internet access. Provides a flag to toggle Goma at build-time rather than needing to modify your config and change it back.

### DIFF
--- a/src/e-build.js
+++ b/src/e-build.js
@@ -31,8 +31,8 @@ function ensureGNGen(config) {
     return runGNGen(config);
 }
 
-function runNinja(config, target, ninjaArgs) {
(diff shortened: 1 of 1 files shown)


ckerr (approved):
Ahhh you should've PRed this a week ago before Zeta knocked out my internet for a couple of days :smile:

dsanders11:
@ckerr, I've actually had it in my local version for a few weeks but it slipped back in the new PR queue. Now you'll have it for next time!

codebytere (approved):

PR #233, opened 2020-11-12
chore: add JSON schema for config files

dsanders11:
Any interest in adding this? I made it for use with my VS Code extension, where JSON schemas can be used to provide functionality when editing JSON files (see GIF below). It's always nice to have a spec for stuff like the config files. I should be able to make the extension load it from `.electron_build_tools`, so if it lives in this repo then everyone can benefit.

evm-config-schema

### DIFF
--- a/evm-config.schema.json
+++ b/evm-config.schema.json
@@ -1,0 +1,126 @@
+{
+  "title": "JSON schema for EVM configuration files",
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "properties": {
+    "defaultTarget": {
+      "description": "Default build target",
+      "type": "string",
+      "enum": [
+        "breakpad",
+        "chromedriver",
+        "electron",
+        "electron:dist",
+        "mksnapshot",
+        "node:headers"
+      ]
+    },
+    "goma": {
+      "description": "Goma mode to use",
+      "type": "string",
+      "enum": [
+        "cache-only",
+        "cluster",
+        "none"
+      ],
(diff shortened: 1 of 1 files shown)


ckerr:
I don't have strong feelings about it either way.

My main concern is that this would be a solution in search of a problem -- e.g. would there be any validation pass to configs that don't pass the schema? And if so, is that something that's desirable, or should we gracefully handle older configs? Or if we _don't_ have validation, does that mean the schema file lands in the repo and just sits there, possibly overlooked when we add/change keys in the codebase?

dsanders11:
> Or if we don't have validation, does that mean the schema file lands in the repo and just sits there, possibly overlooked when we add/change keys in the codebase?

I wasn't considering using it for validation in this repo, more for documentation since currently the schema of the config file isn't documented and could only be discovered by perusing all of the code. For example, `defaultTarget` is used by `e-build.js` but you'd only know about that functionality if you read that file line by line, since it only appears on a single line.

But yes, I see the concern that it might just immediately go stale. Just offering it up since I already made it.

ckerr:
This has been open for a year now.

I still don't have any strong feelings about this PR either way, but if nobody else (either @dsanders11 or another maintainer) gives a shout out to champion for this PR in the next few days, I'm going to go ahead and close it

dsanders11:
@ckerr, I'll go ahead and close it, I don't imagine anything will change in the next few days, and generally I have had little success championing stuff within `build-tools`, so I'm not going to attempt to in this case.

PR #238, opened 2020-12-02
feat: Convert check-for-updates command to an auto-update command with more options

itsananderson:
In order to enable auto updates to be easily enabled/disabled, I've converted
the `check-for-updates` command to an `auto-update` command which can be used to
enable and disable automatic updates in addition to manually triggering an update check.

Auto-updates are enabled/disabled by creating an empty `.disable-auto-updates` file in
the root of the repository. That file is checked before an auto-update is performed, and
if it exists no update check happens.

Updates can still be manually performed with `e auto-update check` or just `e auto-update`.

The old `e check-for-updates` command is kept as an alias for backward compatibility.

If this all seems to complicated, I'm open to closing this PR in favor of a smaller change to just silently check for the existence of a `.disable-auto-updates` file https://github.com/electron/build-tools/compare/master...itsananderson:disable-auto-updates?expand=1


codebytere (approved):
Thanks @itsananderson!

MarshallOfSound (approved):

itsananderson:
👋  do I need to do anything to unblock merging this? Are we just waiting for sign-off from @ckerr?

PR #245, opened 2021-01-16
chore: invoke python2 explicitly when it's present

### LINKED ISSUE #244: build-tools fails when default python version is 3
On systems where `python` invokes python3 rather than python2, build-tools fails:

```sh
Writing new goma.gn file /home/charles/electron/build-tools/third_party/goma.gn
Downloading "https://electron-build-tools.s3-us-west-2.amazonaws.com/build-dependencies/f615bf878f8eeee7bb8bc87034cc22e3a8a5a88e8ad63f45bdd6441979c63b8a/goma-linux.tgz" into /home/charles/electron/build-tools/third_party/goma-linux.tgz
[==================================================================================================================================================] 29444631/bps 100% 0.0s
Not Authenticated - Triggering Goma Login
Running "python goma_auth.py login" in /home/charles/electron/build-tools/third_party/goma
Traceback (most recent call last):
  File "goma_auth.py", line 19, in <module>
    import SocketServer
ModuleNotFoundError: No module named 'SocketServer'
ERROR Failed to run command, exit code was "1", error was 'undefined'
```

The module name in python3 has changed to `socketserver`. It looks like the upstream version of goma_auth.py does not rely on SocketServer or socketserver, so either our version is out-of-date or we've added in some python2 assumptions ourselves. It looks like this code is extracted from the per-archive files sitting on AWS so I don't know its history. @MarshallOfSound do you know if it would be easy to sync the scripts with upstream s.t. they don't depend on a particular version of python?

Setting goma aside, I believe some of Electron's scripts also assume python2. Until the toolchain is all python3-friendly, build-tools should probably invoke python2 explicitly when available.

ckerr:
Fixes #244 by invoking `python2` explicitly when present; `python` otherwise.

A better solution would be to make the toolchain work on both 2 and 3. This PR is an interim fix because my system is broken Right Now and I want it to work :smile:

### DIFF
--- a/src/utils/paths.js
+++ b/src/utils/paths.js
@@ -20,6 +20,6 @@ function ensureDir(dir) {
 
 module.exports = {
   ensureDir,
-  python2: which.sync('python2') || which.sync('python'),
+  python: which.sync('python2') || which.sync('python'),
   resolvePath,
 };
--- a/src/utils/paths.js
+++ b/src/utils/paths.js
@@ -1,6 +1,7 @@
 const fs = require('fs');
 const os = require('os');
 const path = require('path');
+const which = require('which');
 const { color } = require('./logging');
 
 function resolvePath(p) {
@@ -18,6 +19,7 @@ function ensureDir(dir) {
 }
 
 module.exports = {
-  resolvePath,
   ensureDir,
+  python2: which.sync('python2') || which.sync('python'),
+  resolvePath,
 };
--- a/src/e
+++ b/src/e
@@ -187,7 +187,7 @@ program
       goma.downloadAndPrepare(evmConfig.current());
       cwd = goma.dir;
       args[0] = `${args[0]}.py`;
-      args.unshift(paths.python2);
+      args.unshift(paths.python);
     }
 
     const { status, error } = depot.spawnSync(evmConfig.current(), args[0], args.slice(1), {
--- a/src/e
+++ b/src/e
@@ -9,6 +9,7 @@ const evmConfig = require('./evm-config');
 const { color, fatal } = require('./utils/logging');
 const depot = require('./utils/depot-tools');
 const goma = require('./utils/goma');
+const paths = require('./utils/paths');
 const { refreshPathVariable } = require('./utils/refresh-path');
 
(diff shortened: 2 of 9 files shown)


src/utils/paths.js
```diff
@@ -18,6 +19,7 @@ function ensureDir(dir) {
 }
 
 module.exports = {
-  resolvePath,
   ensureDir,
+  python2: which.sync('python2') || which.sync('python'),
```
codebytere, line 23: given that this will not _always_ be python2, we should probably name this just `python`

ckerr: Alright, FWIW my thought was that since we're currently explicitly requiring python2, it's marginally useful to make that explicit in the variable name, and that the variable name could be changed when that dependency changes. I don't feel strongly about either approach though :slightly_smiling_face: 

Changed in 207c8c9b0e55d3753c5d07169d984e7b71778342

ckerr:
Oop, I need to come back to this to fix the Windows issue

ckerr:
Closing this PR because even if this was ready to merge it wouldn't solve the issue. There are third-party scripts that still fail with this PR, e.g. 

```sh
FAILED: gen/third_party/electron_node/src/inspector/js_protocol.json 
python ../../v8/third_party/inspector_protocol/convert_protocol_to_json.py gen/third_party/electron_node/src/inspector/js_protocol.pdl gen/third_party/electron_node/src/inspector/js_protocol.json
Traceback (most recent call last):
  File "../../v8/third_party/inspector_protocol/convert_protocol_to_json.py", line 35, in <module>
    sys.exit(main(sys.argv[1:]))
  File "../../v8/third_party/inspector_protocol/convert_protocol_to_json.py", line 30, in main
    json.dump(protocol, output_file, indent=4, separators=(',', ': '))
  File "/usr/lib/python3.8/json/__init__.py", line 180, in dump
    fp.write(chunk)
TypeError: a bytes-like object is required, not 'str'
[4152/35741] ACTION //third_party/blink/renderer/bindings:generate_bindings_all(//build/toolchain/linux:clang_x64)
```

We can work through these individual problems e.g. by filing upstream tickets, but it still means this PR is not the Right Thing.

In case anyone reads this because they're hit by the same issue -- running `sudo apt install python-is-python2` is brute-force but works.

PR #249, opened 2021-02-04
chore: add support for Microsoft Goma cluster

jkleinsc:
This PR adds support for the Microsoft Goma cluster for those who have access to that cluster.

### DIFF
--- a/src/e
+++ b/src/e
@@ -155,7 +155,14 @@ program
 program
   .command('update-goma')
   .description('Ensure a fresh copy of Goma is installed')
-  .action(() => console.log(`goma checksum: ${goma.downloadAndPrepare()}`));
+  .action(() => {
+    const args = process.argv.slice(3);
+    const config = {};
+    if (args.length > 0 && args[0] === 'msftGoma') {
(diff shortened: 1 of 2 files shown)


src/e
```diff
@@ -155,7 +155,14 @@ program
 program
   .command('update-goma')
   .description('Ensure a fresh copy of Goma is installed')
-  .action(() => console.log(`goma checksum: ${goma.downloadAndPrepare()}`));
+  .action(() => {
+    const args = process.argv.slice(3);
+    const config = {};
+    if (args.length > 0 && args[0] === 'msftGoma') {
```
MarshallOfSound, line 161: Why can't we load the actual config here?

jkleinsc: @MarshallOfSound I wanted a way to download Goma without having to setup a config

codebytere (approved):

ckerr (approved):
I mostly feel like deferring to @MarshallOfSound on this since I don't have a lot of goma domain knowledge; but the PR looks reasonable to me :+1:

MarshallOfSound (approved):

PR #255, opened 2021-02-18
feat: Use goma batch files in Windows

rzhao271:
This PR makes it so that the goma batch files are run instead of the goma python scripts on Windows.
The goma batch files contain logic to select the right Python to use (i.e. vpython, falling back to python if that's not available), and then running the goma python script with that Python. By using the right Python, Windows users can avoid a misleading error about pywin32 not being installed.


codebytere (approved):

rzhao271:
Rebased to fix conflict

PR #256, opened 2021-02-22
perf: set number of Goma subprocs equal to CPUs on Mac

AroundKevin:
Mitigates #229 by setting the number of Goma subprocesses on Mac OS equal to the number of CPUs. Has no effect on other build platforms.

The default on Mac is to use 3/4 subprocesses which is very slow. With this change, using `e build -j 16` on a recent 16" MacBook resulted in ~3x compile time improvement. Current Goma args can be checked at `http://localhost:8088/flagz` with Goma started.

### DIFF
--- a/src/utils/goma.js
+++ b/src/utils/goma.js
@@ -6,6 +6,7 @@ const rimraf = require('rimraf');
 const { unzipSync } = require('cross-zip');
 const { color, fatal } = require('./logging');
 const depot = require('./depot-tools');
+const os = require('os');
 
 const gomaDir = path.resolve(__dirname, '..', '..', 'third_party', 'goma');
 const gomaGnFile = path.resolve(__dirname, '..', '..', 'third_party', 'goma.gn');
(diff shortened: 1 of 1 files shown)


MarshallOfSound (approved):

src/utils/goma.js
```diff
@@ -154,12 +155,23 @@ function ensureGomaStart(config) {
   const { status } = childProcess.spawnSync(gomacc, ['port', '2']);
   if (status === 0) return;
 
+  // Set number of subprocs to equal number of CPUs for MacOS
+  var subprocs = {};
```
MarshallOfSound, line 159:
```suggestion
  let subprocs = {};
```

src/utils/goma.js
```diff
@@ -154,12 +155,23 @@ function ensureGomaStart(config) {
...
+  if (process.platform === 'darwin') {
+    var cpus = os.cpus().length;
```
MarshallOfSound, line 161:
```suggestion
    const cpus = os.cpus().length;
```

PR #274, opened 2021-04-21
chore: use github device auth for github API requests

MarshallOfSound:
Instead of making folks figure out github tokens and environment variables, let's use device-code based auth.


codebytere (approved):
rad

PR #276, opened 2021-04-22
fix: missing notes when cherry-pick can't find bug

codebytere:
Fixes failing check when bug can't be found in Chromium cherry-picks


ckerr (approved):

codebytere:
Closing as we want to force notes if none can be found.

Issue #87, opened 2020-02-15
Initializing in root directory fails to create .gclient config


samuelmaddock:
When initializing the code on Windows, the process fails due to `.gclient` file not existing. I'm going to keep this issue open while I work through fixing it.

```
Sam@blackmesa MINGW64 /d/work/electron/master
$ e init --root=./ --bootstrap testing
New build config testing created in D:\work\build-tools\configs\evm.testing.json
Now using config testing
Creating D:\work\electron\master\src
Running "python D:\work\build-tools\third_party\depot_tools\gclient.py sync --with_branch_heads --with_tags -v --ignore_locks" in D:\work\electron\master\src
Looking for .gclient starting from D:\work\electron\master\src

Couldn't find configuration file.
Error: client not configured; see 'gclient config'
ERROR Error: Command failed: python D:\work\build-tools\third_party\depot_tools\gclient.py sync --with_branch_heads --with_tags -v --ignore_locks
    at checkExecSyncError (child_process.js:630:11)
    at Object.execFileSync (child_process.js:648:15)
    at Object.depotExecFileSync [as execFileSync] (D:\work\build-tools\src\utils\depot-tools.js:86:16)
    at runGClientSync (D:\work\build-tools\src\e-sync.js:35:9)
    at Object.<anonymous> (D:\work\build-tools\src\e-sync.js:48:3)
    at Module._compile (internal/modules/cjs/loader.js:1157:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1177:10)
    at Module.load (internal/modules/cjs/loader.js:1001:32)
    at Function.Module._load (internal/modules/cjs/loader.js:900:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:74:12) 
ERROR Error: Command failed: node D:\work\build-tools\src\e sync -v --ignore_locks      
    at checkExecSyncError (child_process.js:630:11)
    at Object.execFileSync (child_process.js:648:15)
    at Object.<anonymous> (D:\work\build-tools\src\e-init.js:183:18)
    at Module._compile (internal/modules/cjs/loader.js:1157:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1177:10)
    at Module.load (internal/modules/cjs/loader.js:1001:32)
    at Function.Module._load (internal/modules/cjs/loader.js:900:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:74:12) 
[...]

PR #280, opened 2021-04-28
fix: handle root dir existing with no .gclient

codebytere:
Closes https://github.com/electron/build-tools/issues/87.

Cleans up logic a bit to handle the case where `root` has been set to an existing directory but it's not been initialized with a `.gclient` file.

cc @MarshallOfSound @ckerr


ckerr (approved):
Looks good!

PR #313, opened 2021-11-23
feat: enable using build-tools with Chromium

codebytere:
Fixes an issue where running `e sync` with an unaltered Chromium src checkout would fail to sync becuase it would try to set remotes for directories which didn't exist (`src/third_party/electron_node` and `src/electron`)


ckerr (approved):
LGTM

Issue #302, opened 2021-09-21
Allow build-tools to be used with non-standard (e.g. non-github) repositories.


codebytere:
Use case - a fork of Electron in Azure Devops. A user should be able to pass e.g. `e init --repo https://your_repo_url.com/extra/repo/info` and `build-tools` will clone and set up Electron with that instead.

cc @deermichel

codebytere:
Closing as there's not a good path forward here now and this isn't a use case we intend to support in the near-term.

PR #315, opened 2021-11-25
feat: allow build-tools to be used with non-standard (e.g. non-github) repositories

BlackHole1:
closed: https://github.com/electron/build-tools/issues/302

### DIFF
--- a/src/e-init.js
+++ b/src/e-init.js
@@ -31,9 +31,11 @@ function createConfig(options) {
   if (options.tsan) gn_args.push('is_tsan=true');
 
   const electron = {
-    origin: options.useHttps
-      ? 'https://github.com/electron/electron.git'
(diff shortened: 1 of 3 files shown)


src/e-init.js
```diff
@@ -77,7 +79,7 @@ function runGClientConfig(config) {
     '--name',
     'src/electron',
     '--unmanaged',
-    'https://github.com/electron/electron',
+    config.remotes.electron.origin,
```
ckerr, line 82: Does this change still work on the default repo when `useHttps` is not set, i.e. when origin is `git@github.com:electron/electron.git`?

It looks like the `https://` being replaced here was here since the initial commit, so I can't tell from the comment history whether its use was intentional or if it snuck in by accident.

codebytere:
Closing as stale.

PR #316, opened 2021-12-07
fix: detect xcode version after ci config change

clavin:
https://github.com/electron/electron/pull/31741 moved a lot of information in the Circle CI config `config.yml` to a new `build_config.yml` file. The Xcode detection code relied on finding Xcode info in that old file path. This PR introduces the new file path as the source of truth, but still checks the old path if the first one fails to preserve backwards compatibility with electron repos older than that PR. Not sure when the old path should be removed from build tools though.


VerteDinde (approved):
Great catch - thanks so much for catching and fixing this! 🙇‍♀️

PR #317, opened 2021-12-09
fix: build failed

BlackHole1:
see: https://github.com/electron/electron/pull/31741

This PR moves the xcode field information from config.yml to build_config.yml

This caused build-tools to fail to read xcode information, thus downgrading to xcode version 11.1.0

image
image


BlackHole1:
Repeat with https://github.com/electron/build-tools/pull/316, closed

PR #319, opened 2022-01-19
refactor: cleanup e-show

codebytere:
Miscellaneous cleanup discovered when attempting to write a https://fig.io/ config for `build-tools`

### DIFF
--- a/src/e-show.js
+++ b/src/e-show.js
@@ -39,27 +39,18 @@ program.description('Show information about the current build config');
 
 program
   .command('current')
-  .description('Show name of the current build config')
+  .description('Show the current build config')
+  .option('-n, --no-name', "Don't show config name", false)
   .option('-g, --git', 'Human-readable git status (tag, branch, commit)', false)
   .option('-f, --filepath', 'Config filepath', false)
   .action(options => {
     try {
       const name = evmConfig.currentName();
       const parts = [];
-      if (options.name) {
-        const name = color.config(name);
-        parts.push(`Config Name: ${name}`);
-      }
-      if (options.git) {
-        const status = color.git(gitStatus(evmConfig.current()));
-        parts.push(`Git Status: ${status}`);
-      }
-      if (options.filepath) {
-        const path = color.path(evmConfig.pathOf(name));
-        parts.push(`Filepath: ${path}`);
-      }
-
-      const txt = parts.join('\n');
+      if (options.name) parts.push(color.config(name));
+      if (options.git) parts.push(color.git(gitStatus(evmConfig.current())));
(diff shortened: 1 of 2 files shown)


src/e-show.js
```diff
@@ -39,18 +39,27 @@ program.description('Show information about the current build config');
 
 program
   .command('current')
-  .description('Name of the current build config')
-  .option('-n, --no-name', "Don't show config name", false)
```
codebytere, line 43: This was never used.

src/e-show.js
```diff
@@ -195,24 +198,22 @@ const gomaUrl = 'http://localhost:8088/';
 program
   .command('goma')
   .description(`Watch Goma at work at ${gomaUrl}`)
+  .option('--gn', 'Show path of the goma.gn file')
```
codebytere, line 201: I think it's more intuitive to combine these since they're very similar.

src/e-show.js
```diff
@@ -154,29 +163,23 @@ program
...
-    } catch (e) {
-      fatal(e);
-    }
-  });
-
-program
-  .command('outdir')
-  .description('outdir path, e.g. "/$root/src/out/Testing"`')
+  .description('Show outdir name, e.g. "Testing"')
+  .option('--path', 'Output as JSON')
```
codebytere, line 167: I think it's more intuitive to make showing the `outdir` path an option under `out` 🤔

ckerr (approved):
These changes make sense to me.

Note, the `e show gomadir`and `e show gomagn` were originally added for https://github.com/electron/build-tools/pull/129 related to https://github.com/electron/electron/pull/22784 for CI. If this is still in use, then @jkleinsc should be aware of this PR. (If it _isn't_ still in use, maybe this section can just be removed?)

Approving because this code is fine, but also req'ing review from John

jkleinsc (changes requested):
In regards to @ckerr's comment, we never made a change to call `e show gomadir` or `e show gomagn` in CI; we instead directly call:
```
node -e "console.log(require('./src/utils/goma.js').gnFilePath)"
node -e "console.log(require('./src/utils/goma.js').dir)"
```
Marking as request changes as it looks like tests aren't passing

ckerr:
@jkleinsc requesting re-review, I've fixed up the PR.

changes that I pushed:

- removed `e show gomadir` and `e show gomagn` based on John's comment above
- fixed `e show out` 
- updated tests to use the new `--filepath` parameter
- restored `e show current`'s output format and the `--no-name` option so tests can query the config filename

jkleinsc (approved):

PR #320, opened 2022-01-20
feat: expose e patches --list-targets

codebytere:
Allow for listing all patch targets without re-exporting patches.


ckerr (approved):
LGTM. This is a nice improvement

PR #322, opened 2022-01-24
docs: fix root directory default in example

bavulapati:
In the hello world example of build tools, the root directory is mentioned to be defaulting to `~/project/electron`. Whereas the config file generated defaults to `$PWD/electron` .

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ You can run a new Electron build with this command:
 ```sh
 # The 'Hello, World!' of build-tools: get and build `main`
 # Choose the directory where Electron's source and build files will reside.
-# You can specify any path you like; this command defaults to ~/projects/electron.
+# You can specify any path you like; this command defaults to `$PWD/electron`.
 # If you're going to use multiple branches, you may want something like:
 # `--root=~/electron/branch` (e.g. `~/electron-gn/main`)
 e init --root=~/electron --bootstrap testing


codebytere (changes requested):
While we appreciate your efforts, these changes do not meaningfully improve the codebase, and as such we aren't able to accept them at ths time.

README.md
```diff
@@ -29,13 +29,13 @@ You can run a new Electron build with this command:
...
 # Choose the directory where Electron's source and build files will reside.
-# You can specify any path you like; this command defaults to ~/projects/electron.
+# You can specify any path you like; this command defaults to `$PWD/electron`.
 # If you're going to use multiple branches, you may want something like:
 # `--root=~/electron/branch` (e.g. `~/electron-gn/main`)
 e init --root=~/electron --bootstrap testing
 ```
 
-That command's going to run for awhile. While you're waiting, grab a
+That command's going to run for a while. While you're waiting, grab a
```
codebytere, line 38: this is not a typo: https://www.merriam-webster.com/dictionary/awhile

bavulapati: @codebytere Thanks for the quick response. 
`awhile` as a word is not a typo. Its usage here can be improved by removing `for`. or use the phrase `for a while`.

dsanders11:
@codebytere, could you take another look at the latest version of this PR? 🙂 I do think the change fixing `~/projects/electron` is a good one, since the comment isn't currently accurate.

RaisinTen:
The default value for root as suggested in this PR is indeed aligned with what the code currently does:
https://github.com/electron/build-tools/blob/b57b408e1e1d961d036cc37909d6f928f43ac1e8/src/e-init.js#L116-L120

codebytere (approved):

codebytere:
You're right, thanks for the correction! i'll get this merged as soon as CI passes :)

PR #323, opened 2022-02-02
feat: ability to use system xcode

kjvenalainen:
This PR adds the ability to use system version of XCode (installed in `/Applications`) on Mac OS build hosts by setting a new optional root-level directive in the `e config` file.

The new directive as an example:

```
{
  "goma": "none",
  "root": "build/root/dir",
  "xcode": "system",
  ......
```

Valid values are `'system'` or `'default'`. If not present, `'default'` is assumed.

### DIFF
--- a/src/evm-config.js
+++ b/src/evm-config.js
@@ -184,6 +184,14 @@ function sanitizeConfig(name, overwrite = false) {
     changes.push(`replaced superceded 'origin' property with 'remotes' property`);
   }
 
+  if (config.xcode) {
+    if (!['default', 'system'].includes(config.xcode)) {
+      fatal(`Invalid 'xcode' specified in config, options are 'default' or 'system'`);
+    }
+  } else {
+    config.xcode = 'default';
+  }
+
   if (
     config.goma !== 'none' &&
     config.gen &&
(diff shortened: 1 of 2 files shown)


MarshallOfSound (changes requested):
I'm still processing as to whether this is a good idea or not.  Can you explain your use case?

AroundKevin:
> I'm still processing as to whether this is a good idea or not. Can you explain your use case?

Sure. Primarily this is useful when there is business need to build Electron forks with newer XCode versions than standard. For example, to take advantage of new SDKs/features in the newer XCode, or to link against proper headers in an updated SDK.

With current build tools, changing to a new version of XCode requires modifying:

1. `.circleci` for targeting
2. `src/utils/xcode.js` to add a new version number
3. Manually installing the new XCode to the appropriate build dir to prevent `src/utils/XCode.js` from trying to download non-existent files from the build tools mirror

This change allows existing functionality completely intact, but simplifies the above to:

1. Ensure your preferred XCode is installed in the default path (`/Applications/XCode.app`)
2. Add `"xcode": "system"` to the config.

Hope that clarifies. Happy to answer any additional questions or make some changes.

ckerr:
@MarshallOfSound are you still requesting changes on this PR?

I don't feel strongly about it one way or the other, but don't want to leave it in limbo. We should merge or close it

dsanders11:
Closing this as now build tools uses the system Xcode after #617. Thanks for the PR, @kjvenalainen! You were ahead of your time. 😄

PR #329, opened 2022-02-23
feat: allow specifying XCode version to load

codebytere:
Allows specifying a specific version of XCode to load.

Example usage:

```console
$ e load-xcode --version 13.2.1
$ e build --no-load-xcode
```

### DIFF
--- a/src/e-build.js
+++ b/src/e-build.js
@@ -92,6 +92,7 @@ program
   .option('--gen', 'Force a re-run of `gn gen` before building', false)
   .option('-t|--target [target]', 'Forces a specific ninja target')
   .option('--no-goma', 'Build without goma', false)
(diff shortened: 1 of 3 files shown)


src/e-build.js
```diff
@@ -92,6 +92,7 @@ program
   .option('--gen', 'Force a re-run of `gn gen` before building', false)
   .option('-t|--target [target]', 'Forces a specific ninja target')
   .option('--no-goma', 'Build without goma', false)
+  .option('--no-load-xcode', 'Do not load Xcode prior to building', false)
```
jkleinsc, line 95: @codebytere is this option used?  And if so how?

codebytere: @jkleinsc at the moment `e-build` automatically runs `e-load-xcode` with no arguments which would override the version that had just prior been set

PR #347, opened 2022-03-16
chore: pywin32 is no longer needed on CI

dsanders11:
Follow-up to #363 to also remove `pywin32` from CI.


dsanders11:
@jkleinsc, I actually had this PR removing `pywin32` in draft state for a while because I never got around to confirming that it worked fine, and then I did, but didn't have time to come back to the PR and it looks like in the meantime you landed the change. 😄 As such I've changed this PR to only include the change dropping it from CI, which you didn't cover in your PR.

codebytere (approved):

PR #350, opened 2022-03-19
chore: upgrade Jest version

dsanders11:
TL;DR - The version of `fsevents` (being pulled in through Jest) doesn't play nice with Node v17, leading to a `node-gyp` error during install. Bumping the version of Jest to the latest fixes that.


codebytere (approved):
nice find - thanks!

PR #351, opened 2022-03-22
fix: properly handle target argument in `e build`

dsanders11:
Not sure if this is fall out from the commander upgrade (seems plausible given the code's usage of `rawArgs`), but currently providing a target argument to `e build` isn't working since it doesn't map it to the correct name.

This tightens up the different ways a target can be provided and I think handles all cases robustly now.

Came from https://github.com/electron/electron/pull/33355 where it was showing up in the wild.

@codebytere

### DIFF
--- a/src/e-build.js
+++ b/src/e-build.js
@@ -125,14 +125,24 @@ program
         runGNGen(config);
       }
 
-      // collect all the unrecognized args that aren't a target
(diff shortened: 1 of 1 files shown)


codebytere (approved):
Thanks! and yeah - definitely collateral damage from the commander upgrade 😅

PR #359, opened 2022-03-31
fix: make load-xcode work without a config

nornagon:
Makes `e load-xcode` default to the latest version if there's no current
config.


MarshallOfSound (commented):
The fallback very rarely is "correct", it's there just to ensure the build doesn't keel over and die.  I'd prefer if `load-xcode` instead took an optional `--root` or `--electron-src` argument that you could point at your checkout.

But also, why not just make a config that has your `root` configured correctly? The auto-switching of xcode versions is very usefu

nornagon:
> why not just make a config that has your `root` configured correctly

there's no obvious way to do this with an existing checkout.

nornagon:
> The fallback very rarely is "correct"

I basically always work off main, which basically always needs the latest version of Xcode, so I'd say this is not true, at least for me.

MarshallOfSound:
> there's no obvious way to do this with an existing checkout.

Hm, yeah this isn't documented / easy at all.  I do think it's the better way to do what you want to do though.  Maybe a new `e import-checkout` command or something might be appropriate here in the long term.

In the short term you can just copy this config to `~/.electron_build_tools/configs/evm.{name}.json`

```
{
  "goma": "cluster",
  "root": "/Users/{username}/projects/electron/electron",
  "remotes": {
    "electron": {
      "origin": "git@github.com:electron/electron.git"
    },
    "node": {
      "origin": "git@github.com:electron/node.git"
    }
  },
  "gen": {
    "args": [
      "import(\"//electron/build/args/testing.gn\")",
      "import(\"/Users/{username}/.electron_build_tools/third_party/goma.gn\")"
    ],
    "out": "Testing"
  },
  "env": {
    "CHROMIUM_BUILDTOOLS_PATH": "/Users/{username}/projects/electron/electron/src/buildtools",
    "GIT_CACHE_PATH": "/Users/{username}/.git_cache"
  }
}
```

Some paths might need fiddling 😄 

> I basically always work off main, which basically always needs the latest version of Xcode, so I'd say this is not true, at least for me.

Ah yeah I forgot we changed it recently so that the fallback Xcode is always the newest xcode, it still isn't correct and will lead to issues as soon as folks switch branches and goma stops cache hitting

nornagon:
I still think it's preferable behavior to default to latest rather than crash.

nornagon:
whoops wrong button

ckerr:
This PR has been in limbo for ages. I don't feel strongly about this PR either way but lean towards accepting it on the principle of "we shouldn't crash". @MarshallOfSound if we're not going to take this PR, is there a better way at hand to prevent a crash?

dsanders11:
Closing this out as there's no longer a `load-xcode` command after #617.

PR #369, opened 2022-04-25
feat: use the new az storage for build tools things

MarshallOfSound:
Good bye S3 bucket, hello `dev-cdn.electronjs.org.


MarshallOfSound:
The cert is still being issues but this will work Shortly ™️

MarshallOfSound:
Cert issued, works 👍

PR #381, opened 2022-06-10
chore: refactor to use `goma.auth` in `e-build.js`

dsanders11:
Removes duplicated code and brings it into line with the code in `e-init.js`

There's slight changes in behavior here, namely:

- "Not Authenticated - Triggering Goma Login" will no longer be output from `e build`
- For Goma cluster users, `e init` will now fatal error if the Goma auth process exits with a non-zero status code, which would cause it to bail before building when `--bootstrap` is used. I think this is actually a good thing, as previously that case would have marked the login time down suppressing another attempt for 12 hours, and would have started the build with unauthenticated Goma access

### DIFF
--- a/src/utils/goma.js
+++ b/src/utils/goma.js
@@ -154,10 +154,11 @@ function authenticateGoma(config) {
       },
     });
 
-    if (status !== 0) {
-      console.error(
-        `${color.err} Failed to run command, exit code was "${status}", error was '${error}'`,
-      );
+    if (status !== 0) { 
+      let errorMsg = `${color.err} Failed to run command:`;
+      if (status !== null) errorMsg += `\n Exit Code: "${status}"`;
+      if (error) errorMsg += `\n ${error}`;
+      console.error(errorMsg);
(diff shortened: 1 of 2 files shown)


src/utils/goma.js
```diff
@@ -145,14 +145,22 @@ function authenticateGoma(config) {
...
       stdio: 'inherit',
       env: {
         ...process.env,
         AGREE_NOTGOMA_TOS: '1',
       },
     });
+
+    if (status !== 0) {
+      console.error(
+        `${color.err} Failed to run command, exit code was "${status}", error was '${error}'`,
```
codebytere, line 159: Could we do something more like: https://github.com/electron/build-tools/blob/a7e1b97325ea37947db452605406c85b3ca5b9dd/src/e#L219-L224 here?

codebytere (approved):
LGTM with error output nit :)

dsanders11:
Updated with the nicer error message. 🙂

PR #390, opened 2022-06-16
feat: enable rerunning Appveyor builds

codebytere:
As in title. Follow-up to https://github.com/electron/build-tools/commit/6a81d3ac9b3e58edb6d7181ca87fc09bd3412d36, which only enable rerunning CircleCI workflows.


jkleinsc (approved):

ckerr (approved):
cool!

PR #391, opened 2022-06-21
chore: run all Python scripts via python3

### LINKED ISSUE #220: Drop Chocolatey On Windows?
It would be nice to drop Chocolatey on Windows - that was an unpleasant first part of setting up `build-tools` on Windows. Since one of the `build-tools` design philosophies is to encapsulate rather than pollute the global namespace it kind of works against that (although I understand why it was necessary). In my own personal experience it also unnecessarily ended up installing an extra copy of Python - I already had one but the dependency check couldn't find it for some reason despite it being on the path. Chocolatey is also non-trivial [to fully uninstall](https://chocolatey.org/docs/uninstallation), which is another reason I'm not a fan of using it if it can be avoided.

As of the time of this writing, `depot_tools` bootstraps Python on Windows when `gclient` is used, meaning it really isn't necessary to use Chocolatey to install it. The only tricky thing there is it won't be on the path for running scripts in Electron's `package.json`, but there's already issues there where a `depot_tools` environment is needed to run some of those scripts anyway (see #179). Perhaps a solution like `e npx <script>` can be used to work around those issues.

dsanders11:
Use Python 3 in all the places, but still supports `e d python <script>` to minimize breakage for now.

Also closes #220.


dsanders11:
I've tested this on macOS and Windows. Needed a little tweaking for Windows due to the whole `.bat` extension, and I didn't want to spread around `shell: true` to more places than needed.

codebytere (approved):

Issue #412, opened 2022-09-14
`e cherry-pick` fails to process certain Gerrit PR descriptions


ppontes:
The `e cherry-pick` command fails to parse the bug number from the description in certain Gerrit PRs, due to a difference in how the bug number is represented. Example:

https://chromium-review.googlesource.com/c/chromium/src/+/3758469 

This results in the bug number or the CVE nr. (when used with --security) not being added automatically to the cherry-pick PR release notes.

codebytere:
Closed in https://github.com/electron/build-tools/pull/413

PR #413, opened 2022-09-14
fix: `e cherry-pick` adapt the bug nr. extraction regex for new variant

ppontes:
Addresses #412 .

Adapts the regex matching that is used to parse the bug nr. from Gerrit PRs descriptions, to account for a new variant in how the bug nr. is represented (see example in #412):
- Makes the matching case-insensitive.
- Drops the expectation that a bug nr. following "Bug=" contains the word "chromium".


ckerr (approved):
Short and sweet. LGTM

codebytere (approved):

PR #417, opened 2022-09-16
feat: Use DEVELOPER_DIR to set Xcode path without requiring sudo

### LINKED ISSUE #416: Avoid needing `xcode-select` by setting `DEVELOPER_DIR`
The [man pages](https://keith.github.io/xcode-man-pages/xcode-select.1.html#s) for `xcode-select` say that you can switch your active Xcode directory by setting `DEVELOPER_DIR`.

I haven't tried patching `build-tools` to set `DEVELOPER_DIR` yet, but I think that `utils/xcode.js` could set `process.env.DEVELOPER_DIR` instead of calling `xcode-select -s`, which would avoid prompting for `sudo`.

I *have* tried manually setting the `DEVELOPER_DIR` var when invoking `e build` and that does work (and it doesn't prompt for sudo, because `xcode-select -p` also respects the `DEVELOPER_DIR` var). Even if I have a different Xcode path selected, `e build` still gets the correct version and builds successfully.

```
$ xcode-select -p
/Applications/Xcode.app/Contents/Developer

$ DEVELOPER_DIR=~/.electron_build_tools/third_party/Xcode/Xcode.app/Contents/Developer e build
Running "ninja -j 200 electron" in /Users/$user/electron/master/src/out/Testing
[21/21] STAMP obj/electron/electron.stamp
```

itsananderson:
When you work on multiple projects with different Xcode versions, having `e build` invoke `sudo xcode-select -s` all the time get's annoying.

By setting `DEVELOPER_DIR` in `e-build.js` before calling `e-load-xcode.js`, we make `xcode-select -p` return the correct Xcode path, so `e-load-xcode.js` won't attempt to `xcode-select -s` as sudo. Subsequent subprocess calls to actually build Electron will also pick up that `DEVELOPER_DIR` variable, ensuring that the right Xcode version is used.

```
$ xcode-select -p
/Applications/Xcode.app/Contents/Developer

$ e build
Running "e load-xcode --quiet"
Running "ninja -j 200 electron" in /Users/<user>/electron/master/src/out/Testing
[21/21] STAMP obj/electron/electron.stamp
```

If you actually want to persist the selection change, you can still call `e load-xcode` directly.

```
 $ xcode-select -p
/Applications/Xcode.app/Contents/Developer

$ e load-xcode
Setting your Xcode installation to /Users/<user>/.electron_build_tools/third_party/Xcode/Xcode.app, this will require sudo
Password:
SUCCESS

$ xcode-select -p
/Users/<user>/.electron_build_tools/third_party/Xcode/Xcode.app/Contents/Developer
```

Fixes #416

### DIFF
--- a/src/e-build.js
+++ b/src/e-build.js
@@ -9,6 +9,7 @@ const evmConfig = require('./evm-config');
 const { color, fatal } = require('./utils/logging');
 const depot = require('./utils/depot-tools');
 const goma = require('./utils/goma');
+const { XcodePath } = require('./utils/xcode');
 
 function runGNGen(config) {
   depot.ensure();
@@ -85,6 +86,9 @@ program
         ? target === targets.chromium
         : targets.default === targets.chromium;
       if (process.platform === 'darwin' && !isChromium) {
+        // Setting DEVELOPER_DIR lets us override the Xcode path without requiring sudo
+        process.env.DEVELOPER_DIR = XcodePath;
+
         const result = depot.spawnSync(
           config,
           process.execPath,
--- a/src/e-build.js
+++ b/src/e-build.js
@@ -9,7 +9,6 @@ const evmConfig = require('./evm-config');
 const { color, fatal } = require('./utils/logging');
 const depot = require('./utils/depot-tools');
 const goma = require('./utils/goma');
-const { XcodePath } = require('./utils/xcode');
 
 function runGNGen(config) {
   depot.ensure();
@@ -86,9 +85,6 @@ program
         ? target === targets.chromium
         : targets.default === targets.chromium;
       if (process.platform === 'darwin' && !isChromium) {
-        // Setting DEVELOPER_DIR lets us override the Xcode path without requiring sudo
-        process.env.DEVELOPER_DIR = XcodePath;
-
         const result = depot.spawnSync(
           config,
           process.execPath,
--- a/src/utils/depot-tools.js
+++ b/src/utils/depot-tools.js
@@ -75,6 +75,8 @@ function depotOpts(config, opts = {}) {
     // set these defaults that can be overridden via process.env
     PYTHONDONTWRITEBYTECODE: '1', // depot needs it
     DEPOT_TOOLS_METRICS: '0', // disable depot metrics
+    // Circular reference so we have to delay load
+    DEVELOPER_DIR: require('./xcode').XcodePath, // use build-tools version of Xcode
     ...process.env,
     ...platformOpts(),
     ...config.env,


ckerr:
(Possibly related to #323?)

src/e-build.js
```diff
@@ -85,6 +86,9 @@ program
         ? target === targets.chromium
         : targets.default === targets.chromium;
       if (process.platform === 'darwin' && !isChromium) {
+        // Setting DEVELOPER_DIR lets us override the Xcode path without requiring sudo
+        process.env.DEVELOPER_DIR = XcodePath;
```
MarshallOfSound, line 90: Can we set this inside of our `spawn` and `exec` env helper so that it applies for _all_ commands not just `e build`. E.g. `e test` technically uses xcode, as do various `e d *` commands

itsananderson: Good point about setting for other commands besides `e build`. Can you clarify what you mean by `spawn` and `exec` env helper? I found the `runSpecRunner` helper for `e-test`, and the `execFileSync` and `spawnSync` in `depot-tools.js`. Are those the ones you're talking about, or is there some central place I should add this?

I could also probably just set `process.env.DEVELOPER_DIR` in `src/e` but then that would apply super broadly to every subcommand. Is that a good idea, or would there be bad side effects from doing that so broadly?

itsananderson:
> I _think_ this would supersede #323, so CC'ing @MarshallOfSound since he had opinions in that PR

I don't think this PR currently solves the goals of that PR, but it could probably be modified to do so.

One solution would be to allow `DEVELOPER_DIR` variable to be added to your env config, and if it's present, `build-tools` would use that path rather than overriding it with `~/.electron_build_tools/third_party/Xcode/Xcode.app/Contents/Developer`. This way you could force `build-tools` to use a newer version of Xcode that `build-tools` doesn't know about yet. The drawback is that `build-tools` would no longer verify that you're using the right Xcode version, but I guess that's kind of the objective of overriding with your system Xcode.

src/e-build.js
MarshallOfSound: There's a centralized spot for `execFileSync` and `spawnSync` helpers where we generate env vars for sub procs

https://github.com/electron/build-tools/blob/main/src/utils/depot-tools.js#L74-L84

MarshallOfSound:
This doesn't currently supersede #323 and I still think #323 isn't what we want.

itsananderson:
I've moved the `DEVELOPER_DIR` env var to `depot-tools.js` so now it should apply to any child processes spawned via that mechanism.

@MarshallOfSound what would you think of me removing the `xcode-select -s` code from [e-load-xcode.js](https://github.com/electron/build-tools/blob/ed4564f6a436900ea6c648d35795a6f744162c59/src/e-load-xcode.js#L18-L27)? It's redundant with the `DEVELOPER_DIR` logic now, so the only reason I can think of for keeping it is in case someone wants to explicitly call `e load-xcode` to permanently select the build-tools version of Xcode.

The advantage of removing it (or alternatively putting it behind a `--no-select` flag when invoked from `e-build`) is that it would actually make a workaround for #323 possible. Someone who needs to use a bleeding-edge version of Xcode could do so by by specifying a custom `DEVELOPER_DIR` variable in their evm config.

MarshallOfSound (approved):
For now, this makes certain things way easier.  I'd be hesitant to change anything else without a specific need / issue at hand.

PR #419, opened 2022-10-10
chore: remove unused remotes.node in config

dsanders11:
This stopped being used a while ago at this point, but it's still in the code base, and changes the remote for `src/third_party/electron_node` in an Electron checkout when it probably shouldn't.


dsanders11:
Depends on #420.

jkleinsc (approved):

PR #420, opened 2022-10-10
ci: use lts/-1 with actions/setup-node

dsanders11:
v12 is no longer supported, and I want that sweet, sweet optional chaining.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -22,10 +22,10 @@ jobs:
         git config --global core.autocrlf false
         git config --global core.eol lf
     - uses: actions/checkout@v1
-    - name: Use Node.js v12.x
-      uses: actions/setup-node@v1
+    - name: Setup Node.js
+      uses: actions/setup-node@v3
       with:
-        node-version: 12.x
+        node-version: lts/-1
     - name: Cache node_modules
       id: cache-node-modules
       uses: actions/cache@v2
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,10 +27,10 @@ jobs:
     - uses: actions/checkout@v1
       with:
         fetch-depth: 1
-    - name: Use Node.js v14.x
-      uses: actions/setup-node@v1
+    - name: Setup Node.js
+      uses: actions/setup-node@v3
       with:
-        node-version: 14.x
+        node-version: lts/-1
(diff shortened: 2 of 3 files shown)


codebytere (commented):
Node.js v14 is nearly EOL so personally i'd bump to v16!

dsanders11:
@codebytere, done. Ignore the branch name. 😅

jkleinsc (approved):

jkleinsc (changes requested):
Requesting changes to make sure this doesn't land until https://github.com/electron/electron/pull/35396 lands because right now the Appveyor images are using Node 14.

jkleinsc:
- https://github.com/electron/electron/pull/35396 has been merged into main, but we should make sure it gets backported to all active branches before merging this.

dsanders11:
@jkleinsc, I've reworked this PR again. It now uses `lts/-2` instead of specifying a specific Node.js version, which should get us out of the business of updating it manually. At the moment `lts/-2` is Node v14, and in Oct 2023 it'll be Node v16 I believe. Using `lts/-2` means there's a 6-month gap where it will be on an EOL version, but that's no worse than our current status quo here, and it seems like there are concerns about out-running `e/e` so `lts/-1` doesn't seam feasible with that in mind.

Also bumps the minimum Node.js engine version in `package.json` just so we can safely use optional chaining in the codebase.

Thanks @bnb for the tip regarding relative node versions with `actions/setup-node`.

codebytere:
> it seems like there are concerns about out-running e/e so lts/-1 doesn't seem feasible with that in mind.

I'm not sure I follow - out-running in what sense? All supported versions are v16 at least 🤔

codebytere (commented):
https://github.com/electron/electron/pull/35396/files#diff-5021647459797fea2c16a3883972a7ef10723c485325f49465233c936a3851b6R60 appears to have switched from installing `nodejs` to `nodejs-lts` - we shouldn't need to remain 2 behind i don't believe and can likely do `lts/-1` though paging @jkleinsc to confirm!

jkleinsc:
@dsanders11 @codebytere we have merged all the PRs to update active branches to use Node 16 for Appveyor, so we should be good to go with lts-1

jkleinsc (approved):

codebytere (approved):

PR #421, opened 2022-10-10
chore: add JSON schema for config files

dsanders11:
Resurrecting #233, but using the schema to validate during tests.


tests/example-configs-spec.js
```diff
@@ -0,0 +1,21 @@
+const fs = require('fs');
+const path = require('path');
+
+const yml = require('js-yaml');
+
+const { validateConfig } = require('../src/evm-config');
+
+describe('example configs', () => {
+  it('should validate', () => {
+    const exampleConfigsPath = path.resolve(__dirname, '..', 'example-configs');
```
codebytere, line 10: Could we add a test for an invalid config file?

jkleinsc (approved):

donatocayurin:
Thank you ❤️🙏

donatocayurin:
❤️🙏

PR #435, opened 2022-11-22
fix: allow overriding `DEVELOPER_DIR`

codebytere:
Allow users to override `DEVELOPER_DIR`.

Previously, if `DEVELOPER_DIR` was overridden, `e load-xcode` would try to switch it back every time.

### DIFF
--- a/src/e-load-xcode.js
+++ b/src/e-load-xcode.js
@@ -18,12 +18,17 @@ if (Xcode.ensureXcode() === false) {
 // Select our new xcode
 const output = childProcess.execFileSync('xcode-select', ['-p']).toString();
 if (!output.trim().startsWith(Xcode.XcodePath)) {
-  console.info(
-    `Setting your Xcode installation to ${color.path(Xcode.XcodePath)}, this will require sudo`,
-  );
-  childProcess.execFileSync('sudo', ['xcode-select', '-s', Xcode.XcodePath], {
-    stdio: 'inherit',
-  });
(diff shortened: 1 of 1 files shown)


src/e-load-xcode.js
```diff
@@ -18,12 +18,17 @@ if (Xcode.ensureXcode() === false) {
...
-  console.info(
-    `Setting your Xcode installation to ${color.path(Xcode.XcodePath)}, this will require sudo`,
-  );
-  childProcess.execFileSync('sudo', ['xcode-select', '-s', Xcode.XcodePath], {
-    stdio: 'inherit',
-  });
+  if (process.env.DEVELOPER_DIR.length) {
+    console.info(`${color.info} Using overridden Xcode at ${process.env.DEVELOPER_DIR}`);
+    return;
+  } else {
```
MarshallOfSound, line 24:
```suggestion
  if (process.env.DEVELOPER_DIR) {
    console.info(`${color.info} Using overridden Xcode at ${process.env.DEVELOPER_DIR}`);
    return;
  } else {
```

MarshallOfSound (commented):
Should we let `DEVELOPER_DIR` override our xcode? Or should we make people set something like `ELECTRON_BUILD_TOOLS_DEVELOPER_DIR` instead thus ensuring that CI / build systems that already set the standard dev-dir env var don't lose cache hits.

dsanders11:
Closing this as the modified file no longer exists and the concept no longer applies after #617.

PR #436, opened 2022-11-23
feat: add `e shell` command

MarshallOfSound:
This command spawns your current shell with all the environment variables that build-tools normally uses fully in your environment.  Sometimes I find myself running `e d {thing}` a _lot_ and it's annoying not having things like bash completion or the ability to copy paste commands from chromium docs.

This fixes that by spawning a temporary populated shell with everything you need to run things like `ninja` and `gn` directly.  The extra env var is for folks like me who want to be able to customize their shell prompt when running in this mode, see below for an example.

image

I was doing this locally with an alias `SHELL_CONTEXT=" $(e show current)" e d $SHELL` but figured other folks might benefit from it

### DIFF
--- a/src/e
+++ b/src/e
@@ -279,6 +279,33 @@ program
     process.exit(0);
   });
 
+program
+  .command('shell')
+  .description(
+    "Launch a shell environment populated with build-tools' environment variables and context",
+  )
+  .action(() => {
+    depot.ensure();
+
+    if (!['linux', 'darwin'].includes(process.platform)) {
+      fatal(`'e shell' is not supported on non-unix platforms`);
+    }
+
+    if (!process.env.SHELL) {
+      fatal('Could not detect shell to launch');
+    }
+
+    console.info(`Launching build-tools shell with ${color.cmd(process.env.SHELL)}`);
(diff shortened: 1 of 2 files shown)


codebytere (commented):
Can we document this somewhere? Beyond that, LGTM

src/e
```diff
@@ -279,6 +279,33 @@ program
...
+    if (!process.env.SHELL) {
+      fatal('Could not detect shell to launch');
+    }
+
+    console.info(`Launching build-tools shell with ${color.cmd(process.env.SHELL)}`);
+    const { status } = depot.spawnSync(evmConfig.current(), process.env.SHELL, [], {
+      stdio: 'inherit',
+      env: {
+        ...process.env,
+        SHELL_CONTEXT: ` ${evmConfig.currentName()}`,
```
dsanders11, line 303: Can this be replaced with the Unicode escape code, `\ue62e`? And/or, should the unicode character be included at all? Seems specific to your setup.

MarshallOfSound: Most people have this unicode character in their terminals, any of the nerdfonts / powerline fonts have it

dsanders11: Ok. Still would prefer it use the escape code.

codebytere (approved):

PR #456, opened 2023-01-31
fix: use ninja.bat on Windows

dsanders11:
Looks like `ninja.exe` was removed in [the commit that dropped ninja binaries](https://chromium.googlesource.com/chromium/tools/depot_tools/+/8f41177abf2c51b2c0114af9abe9a5cd074d5014). `ninja.bat` was added in Nov 2022.

### DIFF
--- a/src/e-build.js
+++ b/src/e-build.js
@@ -51,7 +51,7 @@ function runNinja(config, target, useGoma, ninjaArgs) {
   depot.ensure(config);
   ensureGNGen(config);
 
-  const exec = os.platform() === 'win32' ? 'ninja.exe' : 'ninja';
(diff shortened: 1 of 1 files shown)


codebytere (approved):

codebytere:
@dsanders11 looks like this is causing some issues:
Screenshot 2023-02-01 at 2 10 10 PM

From https://groups.google.com/a/chromium.org/g/chromium-dev/c/027jM6DLkIk:

> For Developers who build on Windows, instead of calling “ninja.exe”, please call “ninja” to resolve the [depot_tools/ninja.bat](https://crsrc.org/d/ninja.bat) wrapper.

so i'm not sure we should actually be calling `ninja.bat` directly?

PR #462, opened 2023-02-01
feat: add ability to set `target_cpu` in gn extra args

codebytere:
As in title.

### DIFF
--- a/src/e-init.js
+++ b/src/e-init.js
@@ -4,7 +4,7 @@ const childProcess = require('child_process');
 const os = require('os');
 const fs = require('fs');
 const path = require('path');
-const program = require('commander');
+const { program, Option } = require('commander');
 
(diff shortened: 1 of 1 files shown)


src/e-init.js
```diff
@@ -13,6 +13,12 @@ const goma = require('./utils/goma');
 const depot = require('./utils/depot-tools');
 const { checkGlobalGitConfig } = require('./utils/git');
 
+// https://gn.googlesource.com/gn/+/master/docs/reference.md?pli=1#var_target_cpu
```
dsanders11, line 16:
```suggestion
// https://gn.googlesource.com/gn/+/main/docs/reference.md?pli=1#var_target_cpu
```

dsanders11 (approved):

Issue #465, opened 2023-02-02
build-tools runs goma auth on test with goma disabled


codebytere:
As in title. `e test` runs `e build` in order to build node headers, but does not properly pass along goma status. it should.

PR #466, opened 2023-02-02
fix: build node headers without goma when disabled in config

codebytere:
Closes https://github.com/electron/build-tools/issues/465.


jkleinsc (approved):

dsanders11:
This got merged before I could comment (on a call), but I'm not sure I understand the code paths involved here.

If `config.goma === 'none'`, shouldn't `e build` also see that and not use Goma? It feels incorrect that we need to provide `--no-goma` as CLI option when the config has Goma disabled.

@codebytere, do you have repro steps for #465? Specifically, how do I confirm I see the same issue you saw - goma auth pops up?

PR #479, opened 2023-04-12
fix: JSON schema Chromium compatibility

codebytere:
The JSON Schema added in https://github.com/electron/build-tools/issues/421 doesn't take into account a user looking to build Chromium itself.

This fixes that by adding an exception for the required schema objects if the `defaultTarget` is `chrome`.

### DIFF
--- a/evm-config.schema.json
+++ b/evm-config.schema.json
@@ -16,7 +16,8 @@
         "electron",
         "electron:dist",
         "mksnapshot",
-        "node:headers"
+        "node:headers",
+        "chrome"
       ]
     },
     "extends": {
@@ -141,6 +142,19 @@
     {
       "required": [
         "extends"
+      ]
+    },
+    {
+      "properties": {
+        "defaultTarget": {
+          "type": "string",
+          "pattern": "chrome"
+        }
+      },
+      "required": [
+        "defaultTarget",
+        "env",
+        "root"
       ]
     },
     {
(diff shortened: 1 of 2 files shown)


evm-config.schema.json
```diff
@@ -143,6 +144,14 @@
         "extends"
       ]
     },
+    {
+      "properties": {
+        "defaultTarget": {
+          "type": "string",
+          "pattern": "chrome"
+        }
+      }
```
dsanders11, line 153:
```suggestion
      },
      "required": [
        "defaultTarget"
      ]
```

Since `defaultTarget` is normally optional, without requiring it in this `oneOf` schema it will match any schema which is missing `defaultTarget`, which then cause the error of matching multiple `oneOf` schemas.

dsanders11 (commented):
Could you add a `example-configs/evm.chromium.yml` example config? That will also add test coverage since the example configs are validated as part of the tests.

example-configs/evm.chromium.yml
```diff
@@ -0,0 +1,12 @@
+root: /path/to/chromium/
+defaultTarget: chrome
+gen:
+  args:
+    # path to goma for faster builds (https://notgoma.com)
+    - import("/Users/user_name/.electron_build_tools/third_party/goma.gn")
+    - is_debug: false
```
dsanders11, line 7:
```suggestion
    - is_debug = false
```

dsanders11 (approved):

PR #522, opened 2023-11-06
feat: allow disabling `depot_tools` auto-update on sync

codebytere:
Allow disabling automatic depot-tools updating - I hit this on the plane on the way over. We call it ourselves but it's also called in some other cases so I took the marker-file approach that [depot_tools understands](https://chromium.googlesource.com/chromium/tools/depot_tools/+/248aa8ba818b752ef507c73db988b6d32bb90df1/update_depot_tools#31) to cover all bases.

### DIFF
--- a/src/e
+++ b/src/e
@@ -67,7 +67,12 @@ program
   .alias('new')
   .command('sync [gclientArgs...]', 'Get or update source code')
   .command('build [options]', 'Build Electron and other things')
-  .alias('make');
+  .alias('make')
+  .command(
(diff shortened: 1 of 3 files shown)


VerteDinde (approved):

dsanders11:
@codebytere, should https://github.com/electron/build-tools/blob/ddb0c275a841e3f7d6d922e85e3f341fa340c4ad/src/e#L236-L286 have been deleted as part of this change?

PR #527, opened 2023-11-16
fix: send "Auto-updates disabled" messages to stderr, not stdout

hardfalcon:
Outputting to stdout breaks commands like `` cd `e show exec` ``, so use stderr instead.

Fixes: 700187712658e6fbdf521aac1a3ca6e8de8f0f2d ("chore: log info message when auto-update disabled")

### DIFF
--- a/src/e
+++ b/src/e
@@ -17,9 +17,13 @@ refreshPathVariable();
 
 function maybeCheckForUpdates() {
   // skip auto-update check if disabled
+  //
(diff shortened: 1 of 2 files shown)


hardfalcon:
Sure, done. Thanks for the suggestion! :)

dsanders11 (approved):
There are more places we should make this change (we should be better in general about keeping separation between stdout and stderr), but this is a good start.

MarshallOfSound (approved):

PR #530, opened 2023-12-01
build: update deps to clean yarn audit

MarshallOfSound:
Forked `chrome-cookies-secure` to drop the transitive dependency on `request`.  It's one of the main uses of that module, but not what we use it for so easier to drop it in a fork.  Also update node-gyp a decent bit


dsanders11 (approved):

MarshallOfSound:
Test failures are unrelated and being investigated separately

PR #533, opened 2023-12-29
feat: add support for reclient

MarshallOfSound:
This isn't fully ready for public consumption but putting it up now for others to see.  Some missing pieces:
* ~~A more developer friendly way of getting auth onto the machine~~ `e d rbe login`
* ~~Either a more generic hack for https://github.com/bazelbuild/reclient/issues/27 or a real fix upstream~~ Moved TLS termination out of nginx and instead the RBE frontend
* No idea if this even remotely works on windows
* Needs some changes to `e/e` as well https://github.com/electron/electron/pull/40850

### DIFF
--- a/src/download.js
+++ b/src/download.js
@@ -8,7 +8,7 @@ const pipeline = promisify(stream.pipeline);
 
 const { fatal } = require('./utils/logging');
 
-const MB_BYTES = 1025 * 1024;
+const MB_BYTES = 1024 * 1024;
 
 const progressStream = function(tokens) {
   var pt = new stream.PassThrough();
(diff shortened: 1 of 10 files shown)


src/download.js
```diff
@@ -8,26 +8,30 @@ const pipeline = promisify(stream.pipeline);
 
 const { fatal } = require('./utils/logging');
 
-const MB_BYTES = 1025 * 1024;
+const MB_BYTES = 1024 * 1024;
```
VerteDinde, line 11: Just curious, why the bump from 1025 -> 1024 here?

VerteDinde (approved):

src/utils/reclient.js
```diff
@@ -0,0 +1,121 @@
...
+
+const reclientDir = path.resolve(__dirname, '..', '..', 'third_party', 'reclient');
+const reclientTagFile = path.resolve(reclientDir, '.tag');
+const reclientHelperPath = path.resolve(
+  reclientDir,
+  `electron-rbe-credential-helper${process.platform === 'win32' ? '.exe' : ''}`,
+);
+const rbeServiceAddress = 'rbe.notgoma.com:443';
+
+const CREDENTIAL_HELPER_TAG = 'v0.1.0';
```
dsanders11, line 17:
```suggestion
const CREDENTIAL_HELPER_TAG = 'v0.2.0';
```

Newest?

src/download.js
MarshallOfSound: A megabyte is 1024 * 1024, idk why it was 1025 before, that was just wrong afaics

src/utils/reclient.js
MarshallOfSound: The `0.2.0` credential helper didn't publish correctly because of a codesigning issue

PR #534, opened 2024-01-01
ci: validate configs in PRs

dsanders11:
Validate configs to better catch typos and other mistakes which might go unnoticed.


ckerr (approved):
:+1:

PR #537, opened 2024-01-11
feat: use reclient on init instead of goma

MarshallOfSound:
As in title, init should use reclient now

### DIFF
--- a/evm-config.schema.json
+++ b/evm-config.schema.json
@@ -26,6 +26,7 @@
       "minLength": 1
     },
     "goma": {
+      "$comment": "Deprecated, use reclient instead",
       "description": "Goma mode to use",
       "type": "string",
       "enum": [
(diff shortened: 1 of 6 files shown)


dsanders11:
CI is hanging on this, but can't tell if it's legitimate or something flaky, CI on other PRs are being quite slow at the moment as well.

ckerr:
Might not just be CI? `yarn test` is hanging for me in `e-show-spec.js`.

The code changes in this PR look fine, so my guess (w/o having looked at the test) is that the test just needs to be updated

MarshallOfSound:
Could be hanging on a reclient auth request?

dsanders11:
Was indeed hanging on the reclient auth request, so I pushed a commit to skip it during tests.

dsanders11 (approved):

jkleinsc (approved):

Issue #539, opened 2024-01-11
Don't Keep Previous Xcodes Indefinitely


dsanders11:
Xcode takes up a lot of disk space, and currently old versions are never deleted. We should either only keep the latest, or create some reasonable middle ground heuristic like keep the last N. Preferably this behavior could be configurable so a user can opt to only keep the latest to save disk space.

https://github.com/electron/build-tools/blob/cc4f6c3dc8a107a827509b957e10e83ccf8c35ec/src/utils/xcode.js#L239-L240

PR #547, opened 2024-01-18
feat: allow deleting old Xcode versions

codebytere:
Closes https://github.com/electron/build-tools/issues/539.

Adds a new config preference to optionally delete unused Xcode versions to save disk space.

### DIFF
--- a/evm-config.schema.json
+++ b/evm-config.schema.json
@@ -21,7 +21,7 @@
       ]
     },
     "preserveXcode": {
-      "description": "Whether to delete unused Xcode versions",
+      "description": "Preserve the N most recent Xcode versions",
       "type": "number",
       "default": 5
     },
--- a/evm-config.schema.json
+++ b/evm-config.schema.json
@@ -20,6 +20,11 @@
         "chrome"
       ]
     },
+    "preserveXcode": {
+      "description": "Whether to delete unused Xcode versions",
+      "type": "boolean",
+      "default": true
+    },
     "extends": {
       "description": "Name of base config to extend",
       "type": "string",
--- a/evm-config.schema.json
+++ b/evm-config.schema.json
@@ -22,8 +22,8 @@
     },
     "preserveXcode": {
       "description": "Whether to delete unused Xcode versions",
-      "type": "boolean",
-      "default": true
+      "type": "number",
+      "default": 5
     },
     "extends": {
       "description": "Name of base config to extend",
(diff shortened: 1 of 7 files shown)


src/e
```diff
@@ -4,7 +4,6 @@ const childProcess = require('child_process');
 const fs = require('fs');
 const program = require('commander');
 const path = require('path');
-const os = require('os');
```
codebytere, line 7: unused

evm-config.schema.json
```diff
@@ -20,6 +20,11 @@
         "chrome"
       ]
     },
+    "preserveXcode": {
```
MarshallOfSound, line 23: Maybe make it a number, and then preserve the `N` most recent `atime`'ed Xcode folders 🤔

jkleinsc (approved):

evm-config.schema.json
```diff
@@ -20,6 +20,11 @@
...
+      "description": "Whether to delete unused Xcode versions",
+      "type": "number",
+      "default": 5
```
dsanders11, line 26:
```suggestion
      "type": "integer",
      "default": 5,
      "minimum": 1
```

Leverage validation here to prevent any weirdness.

dsanders11 (changes requested):
LGTM overall, just one change to enforce valid values for the config option.

evm-config.schema.json
```diff
@@ -20,6 +20,11 @@
         "chrome"
       ]
     },
+    "preserveXcode": {
+      "description": "Whether to delete unused Xcode versions",
```
dsanders11, line 24:
```suggestion
      "description": "Preserve the N most recent Xcode versions",
```

src/utils/xcode.js
```diff
@@ -86,6 +86,22 @@ function getXcodeVersion() {
...
+    .readdirSync(XcodeDir)
+    .map(xcode => {
+      const xcodePath = path.join(XcodeDir, xcode);
+      const { atime } = fs.statSync(xcodePath);
+      return { name: xcodePath, atime };
+    })
+    .sort((a, b) => b.atime - a.atime);
+
+  const { preserveXcode } = evmConfig.current();
+  for (const { name } of recent.slice(0, preserveXcode)) {
```
MarshallOfSound, line 100: I think this is backwards, the most recent are at the front and we want to preserve them.  So this should be:
```suggestion
  for (const { name } of recent.slice(preserveXcode)) {
```

codebytere: ope, you're right thank you 😶‍🌫️

dsanders11 (approved):

MarshallOfSound (approved):

ckerr (approved):

PR #553, opened 2024-01-30
feat: support custom patches config file location

ckerr:
Changes to `e patches`:

1. Support passing in a custom patches config file instead of always using `electron/patches/config.json`
2. Add support for a `config.json` formatted as an array of objects with `patch_dir` and `repo` properties and an optional `grep` property.

This goes hand-in-hand with https://github.com/electron/electron/pull/41174 for more information; see that PR for more information.

CC @mlaurencin, @jkleinsc

### DIFF
--- a/src/e-patches.js
+++ b/src/e-patches.js
@@ -27,6 +27,7 @@ program
       // build the list of targets
       const targets = {};
       const patchesConfig = options.config;
+      if (!fs.existsSync(patchesConfig)) throw `Config file '${patchesConfig}' not found`;
       const configData = JSON.parse(fs.readFileSync(patchesConfig));
       if (Array.isArray(configData)) {
         for (const target of configData) targets[path.basename(target.patch_dir)] = target;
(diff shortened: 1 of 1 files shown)


src/e-patches.js
```diff
@@ -21,9 +26,13 @@ program
 
       // build the list of targets
       const targets = {};
-      const patchesConfig = path.resolve(config.root, 'src', 'electron', 'patches', 'config.json');
-      for (const [key, val] of Object.entries(JSON.parse(fs.readFileSync(patchesConfig)))) {
-        targets[path.basename(key)] = val;
+      const patchesConfig = options.config;
```
codebytere, line 29: If the config is a custom config passed by the user, we should validate that the file exists at the specified path and throw if it doesn't!

ckerr: SGTM. Added in 0dd17e674ae82cd0ff1d8670f8ae20a83df35ad5

codebytere (approved):

ckerr:
(note: no code changes; force-pushing after Approve due to a rebase to sign my commits)

PR #554, opened 2024-02-01
fix: cherry-pick script

clavin:
A recent change to the cherry-pick script includes a typo which tries to call the number `6` as a function.

I also took the liberty to simplify the code using the [Optional chaining operator `?.`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) which has been stable in Node.js since v14 (4 years ago).


codebytere (approved):

codebytere:
@clavin 

> The base branch requires all commits to be signed.

PR #555, opened 2024-02-09
chore: remove lodash dependency

codebytere:
We only used it in one location, which is trivially replaceable with a pure JS function.


ckerr (approved):
Neat!

PR #556, opened 2024-02-09
refactor: remove rimraf usage

codebytere:
`rimraf` was pulled into Node.js core in v14, and we can use native `fs` methods to achieve the same end, e.g. `rimraf.sync(path)` -> `fs.rmSync(path, { force: true, recursive: true })`

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,8 +47,7 @@
     "jest": "^27.5.1",
     "lint-staged": "^10.0.2",
     "nyc": "^14.1.1",
-    "prettier": "^1.19.1",
-    "rimraf": "^3.0.0"
+    "prettier": "^1.19.1"
   },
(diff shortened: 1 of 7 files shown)


ckerr (approved):

dsanders11 (commented):
Two things:
* Could this be put in a util so that `{ force: true, recursive: true }` doesn't need to be repeated?
* I think there's a chance of slight behavior differences, especially around `EBUSY` on Windows - so just something to keep an eye out for after this merges. I'm not too concerned, I just know that `rimraf` has some extra robustness built-in and I'm not sure how much of that carried over into `fs.rmSync`.

dsanders11 (approved):
LGTM. 👍

PR #557, opened 2024-02-09
refactor: update cherry-pick for new crbug ui

codebytere:
Updates `crbug.js` to compensate for migration to the new Chromium issue tracker.

To determine what endpoint to hit, i dug through the [included buganizer script](https://www.gstatic.com/_/buganizer/_/js/k=buganizer.b.en_US.e8zBqVLSP44.2018.O/am=wA/d=1/rs=AI3Ihhp2ogytZehIP7x8HOr1dtgZcvyAng/m=c) and found the issue fetch endpoint.

From there, i looked at the UI to find the CVE label id: 

Screenshot 2024-02-09 at 12 05 53 PM

and reverse engineered the rest to be able to pull out the CVE again.

### DIFF
--- a/src/utils/crbug.js
+++ b/src/utils/crbug.js
@@ -1,66 +1,79 @@
 const fetch = require('node-fetch');
 const chrome = require('@marshallofsound/chrome-cookies-secure');
-const { color } = require('./logging');
+const { fatal } = require('./logging');
 
-const BASE_URL = 'https://bugs.chromium.org';
-const GET_ISSUE = '/prpc/monorail.Issues/GetIssue';
+const BASE_URL = 'https://issues.chromium.org';
 
-async function getChromeCookies(url) {
-  const cookies = await chrome.getCookiesPromised(url);
+const getPayload = (html, start, end) =>
+  html.substring(html.indexOf(start) + start.length, html.indexOf(end));
 
-  return cookies;
-}
-
-async function getXsrfToken(cookies) {
-  const html = await fetch(BASE_URL, {
+async function getXsrfToken(osid) {
+  const html = await fetch(`${BASE_URL}/issues`, {
     headers: {
-      Cookie: `SACSID=${cookies.SACSID}`,
+      Cookie: `OSID=${osid}`,
+      'Content-Type': 'application/json',
     },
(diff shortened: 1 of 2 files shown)


src/utils/crbug.js
```diff
@@ -1,66 +1,79 @@
...
+      // This API call can sometimes return errant invalid characters at the start of the response.
+      const cleaned = rawJSON.substring(rawJSON.indexOf('['), rawJSON.length);
+      return JSON.parse(cleaned);
+    });
 
   return result;
 }
 
+function parseCveFromIssue(issue) {
+  const CVE_ID = 1223136;
```
ckerr, line 56: I don't know why this is hardcoded here. What is `1223136`?

src/utils/crbug.js
```diff
@@ -1,66 +1,79 @@
...
+
+  const issueData = issue[0][1];
+  const issueMetaData = issueData[issueData.length - 1];
+  const cveData = issueMetaData[2][14].find(d => d[0] === CVE_ID);
```
ckerr, line 60: this code, e.g. `[2][14]` is _really_ specific and needs a code comment explainer. I probably couldn't triage a bug report against this coming in cold. :smile_cat:

ckerr (commented):
I should use this feature more.

src/utils/crbug.js
codebytere: See image in PR body - it's the ID corresponding to the bit of metadata for the CVE. it's visible via the UI and also in the payload:

```json
[
                1223136,
                1363614,
                5,
                null,
                "CVE",
                null,
                null,
                [],
                2
            ]
```

ckerr: I do see the image, sorry for not asking correctly. I get that there's a correlation here, but I don't understand what it means :laughing:

ckerr (approved):
I still find this code a little confusing, but there's only so much we can do here if the upstream payload is undocumented.

Morever, this feature is broken right now so it's a question of not-perfect vs broken. 

So :shrug: I'm fine with this; we can take cleanup / documentation as a followup task if needed.

ckerr (approved):

PR #559, opened 2024-02-16
chore: add support for alternative RBE cluster and helper

jkleinsc:
This PR adds two configs to allow use of an alternative RBE cluster with build tools:

* **reclientHelperPath** - Path to alternative reclient credential helper. When using this option, build-tools will assume that the helper is already properly installed at the specified path.
* **reclientServiceAddress** - Alternative RBE cluster address

### DIFF
--- a/evm-config.schema.json
+++ b/evm-config.schema.json
@@ -57,6 +57,14 @@
         "none"
       ]
     },
+    "reclientHelperPath": {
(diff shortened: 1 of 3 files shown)


ckerr (approved):
The code LGTM from a build-tools perspective, but you know more about reclient than me @jkleinsc :smile_cat:

PR #561, opened 2024-02-25
refactor: try authenticating with codespaces on initial build

codebytere:
Refs https://github.com/electron/electron/pull/41428#issuecomment-1962188175.

Reworks reclient auth a bit to be more similar to goma in that if a user tries to build with reclient enabled, they will be prompted to log in. This also allows us to add some extra logic for codespaces so that we can auth during build instead of in. the oncreate lifecycle command.

### DIFF
--- a/src/utils/reclient.js
+++ b/src/utils/reclient.js
@@ -5,6 +5,7 @@ const tar = require('tar');
 
 const { color, fatal } = require('./logging');
 const { deleteDir } = require('./paths');
+const evmConfig = require('../evm-config');
 
 const reclientDir = path.resolve(__dirname, '..', '..', 'third_party', 'reclient');
 const reclientTagFile = path.resolve(reclientDir, '.tag');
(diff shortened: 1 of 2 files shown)


codebytere:
https://github.com/electron/rbe-credential-helper/releases/tag/v0.3.0 is released so this should be unblocked now!

src/utils/reclient.js
```diff
@@ -106,17 +110,46 @@ function reclientEnv(config) {
...
+  }
+
+  if (isAuthenticated()) return;
+
+  const { status } = childProcess.spawnSync(reclientHelperPath, ['login'], {
+    stdio: 'inherit',
+  });
+
+  if (status !== 0) {
+    if (process.env.CODESPACES) {
```
MarshallOfSound, line 139: I'm not sure this is what we want, on code spaces we should leave reclient enabled, just don't try login during init?

codebytere: meaning you'd prefer we maintain that state? i ended up just doing this as I figured it'd be easier

src/utils/reclient.js
```diff
@@ -14,7 +15,29 @@ const reclientHelperPath = path.resolve(
 );
 const rbeServiceAddress = 'rbe.notgoma.com:443';
 
-const CREDENTIAL_HELPER_TAG = 'v0.2.2';
+const CREDENTIAL_HELPER_TAG = 'v0.3.0';
+
+function getTargetPlatform() {
+  let targetPlatform = null;
+
+  const arch = process.arch === 'arm64' ? 'arm64' : 'amd64';
```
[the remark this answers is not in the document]
codebytere: not sure but i can check

jkleinsc (changes requested):
After https://github.com/electron/electron/pull/41428 landed, I don't think this PR is really necessary.  On codespaces when you run `e build`, it will now prompt:
```
Checking for build-tools updates
Running "git pull origin main --rebase --autostash" in /home/builduser/.electron_build_tools
From https://github.com/electron/build-tools
 * branch            main       -> FETCH_HEAD
build-tools is up-to-date
Downloading "https://dev-cdn.electronjs.org/reclient/credential-helper/v0.2.2/electron-rbe-credential-helper-linux-amd64.tar.gz" into /home/builduser/.electron_build_tools/third_party/reclient.tar.gz
[===================================================================================================================================================================================] 77.57MB/s 100% 0.0s
Authentication Status: Not Authenticated

ERROR You do not have valid auth for Reclient, please run "e d rbe login"
builduser@c60311874b19:/workspaces/gclient/src/electron$ 
```

Which I believe is the preferred behavior

Issue #562, opened 2024-02-26
Output incorrectly piped to console on windows


codebytere:
Running commands with `e d <command>` fails to output to console on Windows.

<details><summary>macOS</summary>
<p>

```
electron on git:roller/chromium/main ❯ e d gn --help                                    3:46PM
Running "gn --help"

Commands (type "gn help <command>" for more help):
  analyze: Analyze which targets are affected by a list of files.
  args: Display or configure arguments declared by the build.
  check: Check header dependencies.
  clean: Cleans the output directory.
  clean_stale: Cleans the stale output files from the output directory.
  desc: Show lots of insightful information about a target or config.
  format: Format .gn files.
  gen: Generate ninja files.
  help: Does what you think.
  ls: List matching targets.
  meta: List target metadata collection results.
  outputs: Which files a source/target make.
  path: Find paths between two targets.
  refs: Find stuff referencing a target or file.

Target declarations (type "gn help <function>" for more help):
  action: Declare a target that runs a script a single time.
  action_foreach: Declare a target that runs a script over a set of files.
  bundle_data: [iOS/macOS] Declare a target without output.
  copy: Declare a target that copies files.
  create_bundle: [iOS/macOS] Build an iOS or macOS bundle.
  executable: Declare an executable target.
  generated_file: Declare a generated_file target.
  group: Declare a named group of targets.
  loadable_module: Declare a loadable module target.
  rust_library: Declare a Rust library target.
  rust_proc_macro: Declare a Rust procedural macro target.
  shared_library: Declare a shared library target.
  source_set: Declare a source set target.
  static_library: Declare a static library target.
  target: Declare a target with the given programmatic type.

Buildfile functions (type "gn help <function>" for more help):
  assert: Assert an expression is true at generation time.
  config: Defines a configuration object.
  declare_args: Declare build arguments.
  defined: Returns whether an identifier is defined.
  exec_script: Synchronously run a script and return the output.
  filter_exclude: Remove values that match a set of patterns.
  filter_include: Remove values that do not match a set of patterns.
  foreach: Iterate over a list.
  forward_variables_from: Copies variables from a different scope.
  get_label_info: Get an attribute from a target's label.
  get_path_info: Extract parts of a file or directory name.
  get_target_outputs: [file list] Get the list of outputs from a target.
  getenv: Get an environment variable.
  import: Import a file into the current scope.
  not_needed: Mark variables from scope as not needed.
  pool: Defines a pool object.
  print: Prints to the console.
[...]

PR #563, opened 2024-02-26
fix: show e-depot-tools command output on Windows

codebytere:
Closes https://github.com/electron/build-tools/issues/562.

### DIFF
--- a/src/e-depot-tools.js
+++ b/src/e-depot-tools.js
@@ -32,31 +32,32 @@ program
       return;
     }
 
+    const current = evmConfig.current();
+
     let cwd;
     if (['goma_ctl', 'goma_auth'].includes(args[0])) {
-      goma.downloadAndPrepare(evmConfig.current());
+      goma.downloadAndPrepare(current);
       cwd = goma.dir;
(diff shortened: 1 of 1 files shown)


ckerr (approved):

MarshallOfSound:
`shell: true` was added to fix python3 issues a while back https://github.com/electron/build-tools/pull/391 we should make sure we aren't making that worse

cc @dsanders11

dsanders11 (changes requested):
I can't seem to repro #562 on my machine - I get the expected output?

`shell: true` is necessary otherwise you have to explicitly put `.bat` extensions to run them. Testing on this branch, I actually can't run `e d gn --help` with these changes:

```shell
Running "gn --help"
ERROR Failed to run command:
 Error: spawnSync gn ENOENT
```

@codebytere, can you provide more environment details that might shine some light on the issue? I'm testing on Windows 10, Node.js 16/18/20 and don't see it.

codebytere:
@dsanders11 sorry i abandoned this but: I was running git bash in Windows 11.

PR #566, opened 2024-03-07
feat: allow building for mas

codebytere:
As in title. This doesn't fully solve the issue though so i'm also going to add another PR to allow changing an existing config more simply

### DIFF
--- a/src/e-init.js
+++ b/src/e-init.js
@@ -28,13 +28,20 @@ function createConfig(options) {
   const gn_args = [`import("//electron/build/args/${options.import}.gn")`];
 
   if (options.reclient !== 'none') {
-    gn_args.push(`use_remoteexec = true`);
+    gn_args.push('use_remoteexec = true');
   }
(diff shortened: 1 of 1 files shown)


src/e-init.js
```diff
@@ -126,6 +127,7 @@ program
   .option('--tsan', `When building, enable clang's thread sanitizer`, false)
   .option('--msan', `When building, enable clang's memory sanitizer`, false)
   .option('--lsan', `When building, enable clang's leak sanitizer`, false)
+  .option('--mas', `Build for the macOS App Store`, false)
```
dsanders11, line 130:
```suggestion
  .option('--mas', 'Build for the macOS App Store', false)
```
Only because you made the same change to the other line. 😅

dsanders11 (approved):

MarshallOfSound (approved):
Seems ok, but should we limit this to darwin?

PR #568, opened 2024-03-10
fix: xcode extraction in Chromium

codebytere:
Fix Xcode extraction when we're using `build-tools` to build Chromium instead of Electron. That information is held in `build/mac_toolchain.py`.


src/utils/xcode.js
```diff
@@ -136,14 +141,25 @@ function expectedXcodeVersion() {
     version = fs.existsSync(baseYaml) && extractXcodeVersion(fs.readFileSync(baseYaml, 'utf8'));
   }
 
-  if (!semver.valid(semver.coerce(version))) {
+  // Finally check build/mac_toolchain.py if we're building Chromium.
+  if (defaultTarget === 'chrome') {
+    const macToolchainPy = path.resolve(root, 'src', 'build', 'mac_toolchain.py');
+    console.log(extractXcodeVersion(fs.readFileSync(macToolchainPy, 'utf8')));
```
MarshallOfSound, line 147: Stray console log

ckerr (approved):
LGTM.

As an aside, should the pre-existing `modernMatch` regex have the same parenthesis as this new code? Looks like the patch version is optional but the `.` that precedes it is not optional

dsanders11 (changes requested):
Should this logic be updated as part of this? Otherwise it seems like `e build` will explicitly skip loading Xcode for a Chromium build.

https://github.com/electron/build-tools/blob/28bf599df3ca3c38ba68c79978f8b73ff0ae9eed/src/e-build.js#L101-L108

Also, I know it would be annoying to plumb all the way down to where it's needed, but I feel like using `defaultTarget` is less than ideal as it doesn't cover the case where a user's default target is not Chromium, and they choose that target, e.g. `e build --target chromium`.

codebytere:
@dsanders11 no - you're right! adapted to address your comments.

src/utils/load-xcode.js
```diff
@@ -9,12 +9,15 @@ const { color, fatal } = require('./logging');
 const Xcode = require('./xcode');
 const evmConfig = require('../evm-config');
 
-function loadXcode(quiet = false) {
+function loadXcode(options) {
```
dsanders11, line 12:
```suggestion
function loadXcode(options = {}) {
```

To handle cases where no options are provided, like here:

https://github.com/electron/build-tools/blob/28bf599df3ca3c38ba68c79978f8b73ff0ae9eed/src/e#L187

dsanders11 (approved):

PR #581, opened 2024-04-18
fix: EINVAL when spawning on Windows

codebytere:
Refs https://github.com/nodejs/node/issues/52554

After the latest security fix on Node.js on Windows, spawning `.cmd` files is blocked unless `{ shell: true }` is used.


dsanders11 (approved):
Minor note, the security change also effects `.bat` files (which is what's being used here). LGTM.

VerteDinde (approved):

Issue #588, opened 2024-05-06
`e sync` command fails due to git root mistmatch (Windows 10)


andrewsml:
this happen by running `e sync` command.
it compares two pathes and the only difference is that C: drive is in uppercase
```
________ running 'python3 build/config/siso/configure_siso.py --rbe_instance projects/rbe-chrome-untrusted/instances/default_instance' in 'c:\workspace\electron_release'
gclient(1967) _WriteFileContents:Writing to file c:\workspace\electron_release\.gclient_previous_sync_commits
ERROR Expected git root to be c:\workspace\electron_release\src\electron but found C:\workspace\electron_release\src\electron
```

### Operating System

Windows 10 22H2

### Expected Behavior

`e sync` succeeds

MarshallOfSound:
I assume we need some kind of `path.normalize` around `cwd` as well here --> https://github.com/electron/build-tools/blob/6cfcc75eac849ef6eec50d99d3da604e82ecc6f8/src/e-sync.js#L21-L23

You can try fiddling with that part of build-tools locally and submit a PR if you can make it work? Build tools is working fine on my windows machine so it's probably something specific to your set up? File system, node version, git version, etc.

andrewsml:
@MarshallOfSound , alright. I made a pull request, thanks

PR #589, opened 2024-05-06
fix: e sync, compare paths with no case (#588)

andrewsml:
Path returned by `git rev-parse --show-toplevel` after normalizing contain C: drive in upper case, so further string comparison fails

### DIFF
--- a/src/e-sync.js
+++ b/src/e-sync.js
@@ -18,7 +18,7 @@ function setRemotes(cwd, repo) {
       .trim(),
   );
 
-  if (gitRoot !== cwd) {
+  if (gitRoot.toLowerCase() !== cwd.toLowerCase()) {
     fatal(`Expected git root to be ${cwd} but found ${gitRoot}`);
   }
 


MarshallOfSound (commented):
I don't think lower casing the path is right here as some filesystems are case sensitive, we probably just want to ensure path paths are fully normalized / resolved.

andrewsml:
@MarshallOfSound , Is it fine to make such a patch only for the Windows platform, like that?
https://github.com/electron/build-tools/pull/589/commits/76d42800d6a59452434d090d88693952e28db310

MarshallOfSound (approved):
Updated this to use a generic approach to filesystem path sensitivity

MarshallOfSound:
@andrewsml This PR is now ready to go but we require PRs to this (and some other Electron repos) to come with [signed commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits). Please set that up, or we can extract this PR into a new one with signed commits from a different auther

codebytere:
Closing as abandoned.

PR #590, opened 2024-05-08
fix: e sync, compare paths with no case on Windows (#588)

andrewsml:
path returned by `git rev-parse --show-toplevel` on Windows may contain uppercase characters, causing further string comparison to fail. so for Windows platform it's better to compare the string with no case.


dsanders11:
Please don't open duplicate PRs.

PR #598, opened 2024-06-05
feat: allow only using macOS SDKs in CI

codebytere:
Allow only using macOS SDKs in CI, so that we don't need to download all of XCode. We can then leverage the `SDKROOT` env var to tell the XCode already in Actions runners which SDK to use.

Xcode SDKS for 14 and 13 are [in Azure](https://portal.azure.com/#view/Microsoft_Azure_Storage/ContainerMenuBlade/~/overview/storageAccountId/%2Fsubscriptions%2F3974e2bb-1561-4b5b-872c-162de614763d%2FresourceGroups%2Felectron-build-tools%2Fproviders%2FMicrosoft.Storage%2FstorageAccounts%2Felectronbuildtools/path/xcode/etag/%220x8DA26F3A767F5EB%22/defaultEncryptionScope/%24account-encryption-key/denyEncryptionScopeOverride~/false/defaultId//publicAccessVal/Blob).

We should eventually extend this for all users but this allows it for the GHA migration in the short term.


src/e-build.js
```diff
@@ -102,7 +103,11 @@ program
       }
 
       if (process.platform === 'darwin') {
-        loadXcode({ target, quiet: true });
+        if (process.env.CI) {
```
VerteDinde, line 106: Do we also want to do this here?:
```suggestion
        if (process.env.CI || options.useSdk) {
```
Not sure if build-tools would try to trigger the Xcode download here if it didn't exist

codebytere: oops yep - thanks!

src/e-build.js
```diff
@@ -102,7 +112,11 @@ program
       }
 
       if (process.platform === 'darwin') {
-        loadXcode({ target, quiet: true });
+        if (process.env.CI || options.useSdk) {
```
jkleinsc, line 115:
```suggestion
        if (options.useSdk) {
```
Same as above

src/e-init.js
```diff
@@ -194,7 +199,11 @@ program
 
       // ensure xcode is loaded
       if (process.platform === 'darwin') {
-        loadXcode(true);
+        if (process.env.CI || options.useSdk) {
```
jkleinsc, line 202:
```suggestion
        if (options.useSdk) {
```
Same as above

jkleinsc (changes requested):
This feature shouldn't turn on automatically for CI.

jkleinsc (approved):

ckerr (approved):

PR #599, opened 2024-06-07
fix: add specific Yarn version to package.json

itsananderson:
## Problem

Several parts of `build-tools` invoke Yarn via `npx yarn` calls. We implicitly assume that the globally installed Yarn version is 1.x, and some of the calls like `npx yarn --prod` fail if you have a newer version of Yarn installed globally.

```
% npx yarn --version
3.6.3
% npx yarn --prod
Unknown Syntax Error: Unsupported option name ("--prod").
```

## Proposed Solution

The Yarn project [recommends](https://yarnpkg.com/blog/release/4.0#installing-yarn) enabling Corepack and configuring a specific Yarn version on a per-project basis using the `packageManager` field in your `package.json`. This replaces their previous recommendation to vendor Yarn into your repo and specify it via a `yarnPath` value in your `.yarnrc`.

Because Corepack is not enabled by default, specifying a `packageManager` value won't actually do anything for most developers. However, since I haven't seen anyone else complaining about this issue, I'm guessing it's not a widespread problem at the moment. This PR specifies a `packageManager` value in `package.json` but doesn't touch anything else, including setup docs. If someone does run into conflicts with their globally installed Yarn version, they should be able to work around the issue by running `corepack enable` which will cause the `packageManager` value to be respected by Corepack's `yarn` shim.

## Alternatives Considered

Another way to solve this would be to update our `npx yarn` calls to explicitly call `npx yarn@1.22.22`, but since we invoke Yarn in several places in both this package and in `build-tools-installer`, we would need to either paste the Yarn version into several places in code, or do some awkward abstraction to import the Yarn version from somewhere.


package.json
```diff
@@ -61,5 +61,6 @@
     "hooks": {
       "pre-commit": "lint-staged"
     }
-  }
+  },
+  "packageManager": "yarn@1.22.22"
```
dsanders11, line 65:
```suggestion
  "packageManager": "yarn@1.22.22+sha1.ac34549e6aa8e7ead463a7407e1c7390f61a6610"
```

This matches [how we have it set over in Forge](https://github.com/electron/forge/blob/38f64e4c9e2c79a3b9ff8447b1a23199415ede97/package.json#L180).

itsananderson: Hmm, is that meant to match the sha1 of the vendored Yarn file? I get something different locally, but perhaps I'm hashing the wrong thing?

```
shasum -a 1 .yarn/releases/yarn-1.22.22.cjs
870faaceb1604f213264cfa69f7db5b620d086b2  .yarn/releases/yarn-1.22.22.cjs
```

dsanders11: The way we pulled it for Forge was using the output of `npm show yarn@1.22.22` - not sure if a different approach is needed in this case with the file being vendored.

```
$ npm show yarn@1.22.22

yarn@1.22.22 | BSD-2-Clause | deps: none | versions: 120
📦🐈 Fast, reliable, and secure dependency management.
https://github.com/yarnpkg/yarn#readme

bin: yarn, yarnpkg

dist
.tarball: https://registry.npmjs.org/yarn/-/yarn-1.22.22.tgz
.shasum: ac34549e6aa8e7ead463a7407e1c7390f61a6610
.integrity: sha512-prL3kGtyG7o9Z9Sv8IPfBNrWTDmXB4Qbes8A9rEzt6wkJV8mUvoirjU0Mp3GGAU06Y0XQyA3/2/RQFVuK7MTfg==
.unpackedSize: 5.3 MB

maintainers:
- byk <ben@byk.im>
- bestander <bestander@gmail.com>
- daniel15 <npm@d.sb>
- arcanis <nison.mael@gmail.com>
- danbuild <build@dan.cx>

dist-tags:
berry: 2.4.3     latest: 1.22.22  legacy: 1.21.1   rc: 1.22.0       

published 3 months ago by arcanis <nison.mael@gmail.com>
```

dsanders11: In this case it might be unnecessary since we're vendoring the file itself.

bavulapati:
Screenshot 2024-06-21 at 11 13 08 AM
A deprecation warning is noticed as well.

itsananderson:
I've updated the PR after further thought/experimentation and discussion with David.

I'm no longer vendoring Yarn directly into the repo, but instead just specifying a `packageManager` field in `package.json`. That way *if* anyone else besides me runs into this, they should be able to fix the issue by enabling Corepack on their machine. If anyone else encounters Yarn compatibility issues that aren't addressed by this, we can always follow up based on that feedback, but this seems like the most benign way to pin the expected Yarn version for now.

dsanders11 (approved):

codebytere (approved):

jkleinsc (approved):

PR #612, opened 2024-07-08
fix: don't use CircleCI config to find Xcode version

codebytere:
Refs https://github.com/electron/electron/pull/42763.

We shouldn't be using the CircleCI config files to fetch the XCode version necessary for build now they've been removed - use `src/build/mac_toolchain.py` in Chromium only. This is also what we're doing for the associated SDK logic.

### DIFF
--- a/src/utils/load-xcode.js
+++ b/src/utils/load-xcode.js
@@ -23,7 +23,7 @@ function loadXcode(options = {}) {
     return;
   }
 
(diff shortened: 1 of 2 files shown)


MarshallOfSound:
Seems legit, now we've been using the SDK stuff in CI for a while maybe we should look at switching over to that anyway 🤔

VerteDinde (approved):

PR #617, opened 2024-07-11
refactor: switch to SDK usage on macOS by default

codebytere:
Refs https://github.com/electron/build-tools/pull/612#issuecomment-2213654463.

We've been running with SDK-only usage in CI successfully for a few weeks now - let's try making it the default moving forward. It's faster, simpler, and takes up less space.


src/evm-config.js
```diff
@@ -217,9 +217,9 @@ function sanitizeConfig(name, config, overwrite = false) {
     changes.push(`fixed invalid property ${color.config('reclient: none')}`);
   }
 
-  if (!('preserveXcode' in config)) {
+  if ('preserveXcode' in config) {
     config.preserveXcode = 5;
```
dsanders11, line 221:
```suggestion
  if ('preserveXcode' in config) {
    delete config.preserveXcode;
```

src/evm-config.js
```diff
@@ -217,9 +217,9 @@ function sanitizeConfig(name, config, overwrite = false) {
...
-    changes.push(`defined ${color.config('preserveXcode')} to default value of 5`);
+    changes.push(`removed ${color.config('preserveXcode')} property`);
   }
```
dsanders11, line 223:
```suggestion
  }
  
  if (config.onlySdk) {
    delete config.onlySdk;
    changes.push(`removed ${color.config('onlySdk')} property`);
  }
```

dsanders11 (commented):
Couple of questions:

- I know SDKs are (significantly) smaller than full Xcode, but do want to keep the concept of `preserveXcode` and apply it to the SDKs to keep them from building up indefinitely over time?
- Will this change require a user on a clean machine install Xcode before using `build-tools`? Will they get a reasonable error message that makes it clear what the issue is?
- If we're leaving management of Xcode to users, are we going to run into future issues where they're on too outdated of an Xcode and need to update? Can we detect that and give a proper error message?

Seems like we may want to keep `src/utils/xcode.js` and refactor some of the existing checks (`expectedXcodeVersion()`, `ensureXcode()`, etc) to deal with the above concerns and ensure the system Xcode is what is needed. We could pick up some tricks from #323.

codebytere:
@dsanders11 @MarshallOfSound updated this and addressed some feedback:

- I know SDKs are (significantly) smaller than full Xcode, but do want to keep the concept of preserveXcode and apply it to the SDKs to keep them from building up indefinitely over time? **Done!**
- Will this change require a user on a clean machine install Xcode before using build-tools? Will they get a reasonable error message that makes it clear what the issue is? **Addressed by https://github.com/electron/build-tools/pull/653**
- If we're leaving management of Xcode to users, are we going to run into future issues where they're on too outdated of an Xcode and need to update? Can we detect that and give a proper error message? **Potentially, but if possible i'd rather address that as a follow up and get this tested a little more now**

src/evm-config.js
```diff
@@ -215,9 +215,19 @@ function sanitizeConfig(name, config, overwrite = false) {
...
+    changes.push(`removed ${color.config('preserveXcode')} property`);
+  }
+
+  if (config.onlySdk) {
+    delete config.onlySdk;
+    changes.push(`removed ${color.config('onlySdk')} property`);
+  }
+
+  if (!('preserveSDK' in config)) {
+    config.preserveSDK = 5;
```
dsanders11, line 229: How about moving this block to before removing `preserveXcode` and set the default like `config.preserveSDK = config.preserveXcode ?? 5`? That way if a user has already customized `preserveXcode`, we migrate that preference.

codebytere:
@MarshallOfSound done in [83cd723](https://github.com/electron/build-tools/pull/617/commits/83cd72375e5d6ee78c77fc98e9936920815fc1f2)

MarshallOfSound (approved):

PR #619, opened 2024-07-12
feat: streamline using Reclient for Chromium

codebytere:
Enables building stock Chromium more simply.

Patches are applied now upon sync when targeting Chromium, with reduced area for failure with pinned `EngFlow/reclient-configs` sha.


dsanders11:
Has this been tested on Windows? I have a feeling we want to use `depotSpawnSync` to ensure that the `depot_tools` managed Python 3 gets used there.

codebytere:
@dsanders11 no, i've only done this on posix so far! good call though, i think you're right 😅

VerteDinde (approved):

PR #622, opened 2024-07-25
fix: path comparison in e sync

codebytere:
Supersedes https://github.com/electron/build-tools/pull/589


ckerr (approved):

MarshallOfSound (commented):
```
> path.relative('/Users/', '/users/')
'../users'
```

I don't think this does what we want 🤔

codebytere:
- [ ]

PR #623, opened 2024-07-29
fix(evm-config): import `lodash`

erickzhao:
We're using `_.mergeWith` in this file without importing it.

https://github.com/erickzhao/build-tools/blob/8c95e2259a1085362b6e78fdffb88480c28d6e49/src/evm-config.js#L144-L148


ckerr (approved):

codebytere (approved):
lgtm for now but i think we should probably look to remove lodash entirely

PR #625, opened 2024-08-01
test: switch to vitest

dsanders11:
Significantly reduces CI time on Windows.

Requires bumping the minimum Node.js version to 18 but `e/e` is already on 20 effectively so that shouldn't really matter.


ckerr:
CC @MarshallOfSound regarding Node bump

ckerr (approved):
It's kind of crazy what this does to yarn.lock :boom:

codebytere (commented):
@dsanders11 are there any potential downsides to be aware of here? Was the primary motivating factor CI run time?

dsanders11:
@codebytere, no potential downsides that I'm aware of, other than the Node.js bump (which isn't really a downside).

The primary motivator for opening the PR is indeed the impact on CI time for Windows. In general I've found Vitest to be faster than Jest (not usually this drastic), and it's better suited for the future as it Just Works with ESM, while Jest continues to struggle with ESM support. That's not an issue here at the moment, but it's nice to be on a test framework that won't potentially trip us up long-term.

codebytere (approved):

PR #635, opened 2024-09-05
docs: improve e build --target documentation

codebytere:
This PR adds some improvements to the `e build` documentation to clarify passing `--target` as well as how to see/use possible build targets.

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,7 +207,7 @@ Some of the build targets you may want to build include:
 | electron:node_headers  | Builds the node headers `.tar.gz` file          |
 | electron:electron_symbols  | Generate the breakpad symbols in release builds     |
 
-To build a specific target, run `e build --target [target]`. Running `e build` with no
+To build a specific ninja target, run `e build --target [target]`. Running `e build` with no
 target will build `electron` by default.
 
 Any extra args are passed along to [ninja][ninja], so for example
--- a/README.md
+++ b/README.md
@@ -195,7 +195,7 @@ $ e build --help
 ```
 
 Once you have the source, the next step is to build it with `e build [target]`.
-These build targets are supported:
+Some of the build targets you may want to build include:
 
 | Target        | Description                                              |
 |:--------------|:---------------------------------------------------------|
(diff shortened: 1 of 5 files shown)


VerteDinde (approved):

README.md
```diff
@@ -204,11 +204,44 @@ These build targets are supported:
 | electron      | Builds the Electron binary **(Default)**                 |
 | electron:dist | Builds the Electron binary and generates a dist zip file |
 | mksnapshot    | Builds the `mksnapshot` binary                           |
-| node:headers  | Builds the node headers `.tar.gz` file                   |
+| electron:node_headers  | Builds the node headers `.tar.gz` file          |
+| electron:electron_symbols  | Generate the breakpad symbols in release builds     |
+
+To build a specific target, run `e build --target [target]`. Running `e build` with no
+target will build `electron` by default.
```
dsanders11, line 211:
```suggestion
To build a specific ninja target, run `e build --target [target]`. Running `e build` with no
target will build `electron` by default.
```

README.md
```diff
@@ -204,11 +204,44 @@ These build targets are supported:
...
+
+Any extra args are passed along to [ninja][ninja], so for example
+`e build -v` runs a verbose build.
+
+To see an exhaustive list of all possible build targets, you can run `e d gn ls out/[TYPE]`,
+where `[TYPE]` is e.g. `Testing` or `Release` depending on your build type. This will log a long
+list of targets to the console and also allow you to build some of Chromium's targets.
```
dsanders11, line 218: Worth noting that command must be run from `$root`, otherwise it will error.

README.md
```diff
@@ -204,11 +204,44 @@ These build targets are supported:
 | electron      | Builds the Electron binary **(Default)**                 |
 | electron:dist | Builds the Electron binary and generates a dist zip file |
 | mksnapshot    | Builds the `mksnapshot` binary                           |
-| node:headers  | Builds the node headers `.tar.gz` file                   |
+| electron:node_headers  | Builds the node headers `.tar.gz` file          |
+| electron:electron_symbols  | Generate the breakpad symbols in release builds     |
```
dsanders11, line 208: I think this might cause some confusion.

Unfortunately the semantics between `e build [target]` and `e build --target [target]` are kind of confusing. They're both called `[target]` but in the former it's really just some aliases we define to ninja targets, and in the latter it's ninja targets.

This leads to weirdness like `e build mksnapshot` is building `electron:electron_mksnapshot_zip` while `e build --target mksnapshot` is actually building `v8:mksnapshot`.

Currently the table only lists the aliases, which aligns with the sentence that proceeds it about using `e build [target]`. These changes here introduce ninja targets into the table, which *kind of* work with `e build [target]` but arguably shouldn't. If you run `e build v8:mksnapshot` the output indicates that `autoninja v8:mksnapshot -j 200 electron` is being run, which seems like unexpected behavior and a bug.

dsanders11 (changes requested):
I agree with the overall PR, just calling out some weirdness that we should better document or fix.

ckerr (approved):
LGTM, the only changes I'd suggest have already been covered by @dsanders11

README.md
codebytere: @dsanders11 thinking about it - it's ultimately probably easier to do away with the aliases since the average use won't need them and it's better to be explicit in e.g. ci 🤔 What do you think?

ckerr: I like that idea; I'm fine with removing the aliases

ckerr: I like that idea; I'm fine with removing the aliases.

@dsanders11 ping

README.md
codebytere: This isn't true anymore - i fixed that a bit back so `e build` can be run from anywhere and will run from `$root` on its own :D

src/utils/load-xcode.js
```diff
@@ -10,7 +10,6 @@ const Xcode = require('./xcode');
 const evmConfig = require('../evm-config');
 
 function loadXcode(options = {}) {
-  const target = options.target || 'electron';
```
codebytere, line 13: unused

src/e-build.js
```diff
@@ -122,37 +113,8 @@ program
...
-        if (target) {
-          ninjaArgs.unshift(target);
-        }
-        target = targets['default'];
-      }
-
-      try {
-        runNinja(config, target, ninjaArgs);
-      } catch (ex) {
-        if (target === targets['node:headers']) {
```
codebytere, line 145: No longer an issue for any supported version of Electron

ckerr (approved):
Reapproving the new changes & the alias removal

jkleinsc (approved):

PR #639, opened 2024-09-17
chore: bump engflow-reclient-configs

samuelmaddock:
bumps to https://github.com/EngFlow/reclient-configs/commit/7851c9387a770d6381f4634cb293293d2b30c502

### DIFF
--- a/src/utils/setup-reclient-chromium.js
+++ b/src/utils/setup-reclient-chromium.js
@@ -46,7 +46,7 @@ function configureReclient() {
 
     // Pinning to prevent unexpected breakage.
     const ENGFLOW_CONFIG_SHA =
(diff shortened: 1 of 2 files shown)


MarshallOfSound (approved):
This is only for Chromium builds, you want to update https://github.com/electron/electron/blob/main/DEPS#L16-L17 as well

samuelmaddock:
> This is only for Chromium builds, you want to update https://github.com/electron/electron/blob/main/DEPS#L16-L17 as well

Yep, I fixed this to get Chromium builds working 😄 I'll have to look into how we use it in Electron!

codebytere (approved):

codebytere:
@samuelmaddock can you please sign your commit?

codebytere (approved):

Issue #642, opened 2024-09-18
`e cherry-pick` does not sign commits with PATs


MarshallOfSound:
As in title, we need signed commits now, so uh, change this 🥳

PR #643, opened 2024-09-19
refactor: rewrite cherry-pick command to account for signing requirement

codebytere:
Closes https://github.com/electron/build-tools/issues/642

Rewrites `e cherry-pick` to account for Electron's commit signing requirement. It's not currently possible to sign commits via API unless you're a GitHub App, so we unfortunately need to now temporarily clone e/e and delete it later.

See https://github.com/electron/electron/pull/43799 for successful example!

Screenshot 2024-09-19 at 12 21 48 PM

### DIFF
--- a/src/e-cherry-pick.js
+++ b/src/e-cherry-pick.js
@@ -4,11 +4,20 @@ const d = require('debug')('build-tools:cherry-pick');
 const program = require('commander');
 const https = require('https');
 const got = require('got');
+const cp = require('child_process');
+const fs = require('fs');
+const path = require('path');
+const os = require('os');
 const { Octokit } = require('@octokit/rest');
 
 const { getCveForBugNr } = require('./utils/crbug');
 const { getGitHubAuthToken } = require('./utils/github-auth');
 const { fatal, color } = require('./utils/logging');
+
+const ELECTRON_REPO_DATA = {
+  owner: 'electron',
+  repo: 'electron',
+};
 
 const gerritSources = [
   'chromium-review.googlesource.com',
@@ -131,16 +140,20 @@ program
       patchUrlStr = targetBranch;
       targetBranch = tmp;
     }
+
     const octokit = new Octokit({
       auth: await getGitHubAuthToken(['repo']),
     });
+
+    const tmp = fs.mkdtempSync(path.join(os.tmpdir(), 'electron-tmp'));
+    const electronPath = path.join(tmp, 'electron');
+
+    let error = null;
     try {
       const {
         data: { permissions },
-      } = await octokit.repos.get({
-        owner: 'electron',
-        repo: 'electron',
-      });
+      } = await octokit.repos.get(...ELECTRON_REPO_DATA);
+
       if (!permissions?.push) {
         fatal(
           'The supplied $GITHUB_TOKEN does not have write access to electron/electron - exiting',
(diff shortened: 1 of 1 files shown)


ckerr (approved):

VerteDinde (approved):

src/e-cherry-pick.js
```diff
@@ -157,90 +167,67 @@ program
       const patchPath = `patches/${patchDirName}`;
       const targetBranches = [targetBranch, ...additionalBranches];
 
+      d(`Cloning electron/electron to ${tmp}`);
+      cp.execSync('git clone https://github.com/electron/electron', {
```
MarshallOfSound, line 171: Can we add a log for cloning? And enable stdio, some folks need stdio for ssh passphrases

src/e-cherry-pick.js
```diff
@@ -131,16 +135,22 @@ program
       patchUrlStr = targetBranch;
       targetBranch = tmp;
     }
+
     const octokit = new Octokit({
```
MarshallOfSound, line 139: I think we can drop octokit completely from this flow, we only use it to check permissions now which is not required.

We should load the clone url from `config.remotes.electron.origin` and then if the config is using a fork we'll just push to that (`config.remotes.electron.fork`) instead.

Then instead of using octokit to make a PR (because this device code workflow sucks) we just open them to the pull request create page in the same way `e pr` does.

Then the whole thing works for forks (yay) and with no github auth dance (yay)

src/e-cherry-pick.js
```diff
@@ -131,143 +138,100 @@ program
...
+    const { remotes } = current();
+
+    try {
       const { patchDirName, shortCommit, patch, bugNumber, cve } = await getPatchDetailsFromURL(
         patchUrlStr,
         security,
       );
 
-      const patchName = `cherry-pick-${shortCommit}.patch`;
+      const patchName = `cherry-pick-${shortCommit}1.patch`;
```
MarshallOfSound, line 153: I assume this was left in for testing purposes? Can clean it up?

src/e-cherry-pick.js
```diff
@@ -131,143 +138,100 @@ program
...
-          issue_number: pr.number,
-          labels: [
-            target,
-            'backport-check-skip',
-            'semver/patch',
-            ...(security ? ['security 🔒'] : []),
-          ],
-        });
+        open(
+          `${remotes.electron.origin}/compare/${comparePath}?${querystring.stringify(queryParams)}`,
```
MarshallOfSound, line 224: I don't think this will work for ssh origins?

src/e-cherry-pick.js
```diff
@@ -131,143 +138,100 @@ program
...
 
-        console.log(`Created cherry-pick PR to ${target}: ${pr.html_url}`);
+        d(`Removing temporary electron directory at ${tmp}`);
+        fs.rmSync(tmp, { recursive: true, force: true });
       }
     } catch (err) {
-      console.error('Failed to cherry-pick');
+      console.error(`${color.err} Failed to cherry-pick`);
+
+      fs.rmSync(tmp, { recursive: true, force: true });
```
MarshallOfSound, line 234: Can we move this from `catch` to `finally`

src/e-cherry-pick.js
```diff
@@ -131,143 +138,100 @@ program
...
-          ],
-        });
+        open(
+          `${remotes.electron.origin}/compare/${comparePath}?${querystring.stringify(queryParams)}`,
+        );
+        console.log(`${color.success} Created cherry-pick PR to ${target}`);
 
-        console.log(`Created cherry-pick PR to ${target}: ${pr.html_url}`);
+        d(`Removing temporary electron directory at ${tmp}`);
+        fs.rmSync(tmp, { recursive: true, force: true });
```
MarshallOfSound, line 229: Doing this inside the loop will break multi backport because we clone outside the loop but delete inside. Let's just replace this with a spawn of `git clean -fdx` instead to ensure the checkout is clean between runs of the loop

src/e-cherry-pick.js
codebytere: yes, sorry!

MarshallOfSound (approved):

PR #655, opened 2024-10-11
fix: handle outdated system Xcode version

codebytere:
Addresses https://github.com/electron/build-tools/pull/617#pullrequestreview-2175557963.

Throws a descriptive error if the system Xcode version is too old. We should make this more dynamic to the current OS version but this should be enough for now.

### DIFF
--- a/src/utils/sdk.js
+++ b/src/utils/sdk.js
@@ -119,11 +119,21 @@ function expectedSDKVersion() {
   return version;
 }
 
+// Ensure that the user has a version of Xcode installed and usable.
 function ensureViableXCode() {
   const xcodeBuildExec = '/usr/bin/xcodebuild';
   if (fs.existsSync(xcodeBuildExec)) {
     const result = cp.spawnSync(xcodeBuildExec, ['-version']);
-    if (result.status === 0) return;
+    if (result.status === 0) {
+      const match = result.stdout
+        .toString()
+        .trim()
+        .match(/Xcode (\d+\.\d+)/);
+      if (match && !semver.satisfies(semver.coerce(match[1]), '>14')) {
+        fatal(`Xcode version ${version} is not supported, please upgrade to Xcode 15 or newer`);
+      }
+      return;
+    }
   }
 
   fatal(`Xcode appears to be missing, you may have Command Line Tools installed but not a full Xcode. Please install Xcode now...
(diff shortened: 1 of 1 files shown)


ckerr (approved):

dsanders11 (changes requested):
I think L127 is going to return before the new code gets executed? Could fix that by wrapping the new code inside that if block and return on the branch which has a correct version.

src/utils/sdk.js
```diff
@@ -119,11 +119,21 @@ function expectedSDKVersion() {
...
     if (result.status === 0) return;
+
+    const match = result.stdout
+      .toString()
+      .trim()
+      .match(/Xcode (\d+\.\d+)/);
+    const version = match ? match[1] : 0.0;
+    if (!semver.satisfies(version, '>14')) {
+      fatal(`Xcode version ${version} is not supported, please upgrade to Xcode 15 or newer`);
+    }
```
dsanders11, line 136:
```suggestion
    if (match && !semver.satisfies(semver.coerce(match[1]), '>14')) {
      fatal(`Xcode version ${version} is not supported, please upgrade to Xcode 15 or newer`);
    }
```

Looks like the version output can be formatted like '16.0', which needs to be coerced to a semver version.

Other suggestion in this change is to fall through to the fatal error below rather than using a fake `0.0` value, since the fatal error below is more descriptive.

src/utils/sdk.js
```diff
@@ -119,11 +119,21 @@ function expectedSDKVersion() {
...
-    if (result.status === 0) return;
+    if (result.status === 0) {
+      const match = result.stdout
+        .toString()
+        .trim()
+        .match(/Xcode (\d+\.\d+)/);
+      if (match && !semver.satisfies(semver.coerce(match[1]), '>14')) {
+        fatal(`Xcode version ${version} is not supported, please upgrade to Xcode 15 or newer`);
+      }
+      return;
```
dsanders11, line 135:
```suggestion
      if (match) {
        if (!semver.satisfies(semver.coerce(match[1]), '>14')) {
          fatal(`Xcode version ${match[1]} is not supported, please upgrade to Xcode 15 or newer`);
        } else {
          return;
        }
      }
```

I led things slightly astray with my previous comment, but I think this cleans it up. We want to fall-through to the more detailed error message below if there's no `match`, if the version number either fails to satisfy the check we give the more specific error, else everything is as expected and return.

dsanders11 (approved):

PR #658, opened 2024-10-15
chore: update rbe helper to latest version

jkleinsc:
- Followup to #657; more permanent fix for using read only cache for fork PRs.

### DIFF
--- a/src/utils/reclient.js
+++ b/src/utils/reclient.js
@@ -14,7 +14,7 @@ const reclientHelperPath = path.resolve(
 );
 const rbeServiceAddress = 'rbe.notgoma.com:443';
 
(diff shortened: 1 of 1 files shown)


VerteDinde (approved):

src/utils/reclient.js
```diff
@@ -126,8 +126,6 @@ function reclientEnv(config) {
       const extraArgs = JSON.parse(result.stdout.toString());
       reclientEnv = Object.assign(reclientEnv, extraArgs);
       // Temp fix until reclient helper can be fixed
```
dsanders11, line 128:
```suggestion
```

dsanders11 (changes requested):
Need to drop the dangling comment.

PR #667, opened 2024-10-18
refactor: replace got usage with builtin fetch

dsanders11:
This is pretty straightforward, only `src/download.js` required more involved changes.


ckerr (commented):
:tada:

ckerr (approved):

PR #674, opened 2024-11-01
fix: allow disable download windows toolchain on windows platform

reitowo:
I think it is reasonable for _**local development only**_ that use local VS instead of downloading a huge blob from electron's cdn.

This indeed benefits binary consistency and CI, but seems not important for contributors to develop locally, so perhaps we can disable it by set `ELECTRON_DEPOT_TOOLS_WIN_TOOLCHAIN` to `0`, since this feature is kind of hidden, the developer knows what they are doing.

Even in [official docs](https://www.electronjs.org/docs/latest/development/build-instructions-gn#gn-prerequisites) wants contributing developers to use local toolchain instead of downloading, so I think it should be all right to disable here.

### DIFF
--- a/src/utils/depot-tools.js
+++ b/src/utils/depot-tools.js
@@ -57,7 +57,7 @@ function platformOpts() {
   let opts = {};
 
   const winToolchainOverride = process.env.ELECTRON_DEPOT_TOOLS_WIN_TOOLCHAIN;
-  if (os.platform() === 'win32' || winToolchainOverride === '1') {
+  if ((os.platform() === 'win32' && winToolchainOverride !== '0') || winToolchainOverride === '1') {
     opts = {
       DEPOT_TOOLS_WIN_TOOLCHAIN: '1',
       DEPOT_TOOLS_WIN_TOOLCHAIN_BASE_URL: 'https://dev-cdn.electronjs.org/windows-toolchains/_',


ckerr (approved):
I'm fine with this as a hidden opt-out feature -- it seems harmless enough and could be useful.  But I would like a 2nd opinion

dsanders11 (approved):
Agree with @ckerr, it seems reasonable as an opt-in for developers who specifically want it.

jkleinsc (approved):

MarshallOfSound:
> Even in [official docs](https://www.electronjs.org/docs/latest/development/build-instructions-gn#gn-prerequisites) wants contributing developers to use local toolchain instead of downloading, so I think it should be all right to disable here.

These docs are slightly out of date and note at the [top](https://www.electronjs.org/docs/latest/development/build-instructions-gn#build-tools) that using build-tools automates everything for you, the rest of the page is the "set it up manually" stuff. People _should_ be using the the windows toolchain in order to ensure consistency with our CI environment and other developers.

If we're leaving this in can someone follow up with a PR that disables reclient completely if this flag is set, there's no point spamming Electrons backend with reclient requests if you're not using the toolchain as there's a 0% chance you get a cache hit for anything. cc @electron/wg-infra

reitowo:
That is true. In fact I already set reclient to none, because of pool internet connection.

reitowo:
@MarshallOfSound Take a look at #676

PR #679, opened 2024-11-15
feat: e pr download-dist <number>

samuelmaddock:
`e pr download-dist <number>` will download a pull request's dist artifacts to \
`~/.electron_build_tools/artifacts/pr_{number}_{commithash}_{platform}_{arch}`

As a followup, we could add a `electron-fiddle://` command to register it as a local build for easier testing. With that, this command could register the downloaded artifact.


src/e-pr.js
```diff
@@ -188,5 +188,143 @@ program
...
+
+    // Extract dist.zip
+    await extractZip(distZipPath, { dir: prDir });
+
+    // Check if Electron.app exists within the extracted dist.zip
+    const electronAppPath = path.join(prDir, 'Electron.app');
+    if (!(await fs.promises.stat(electronAppPath).catch(() => false))) {
+      fatal(`Electron.app not found within the extracted dist.zip.`);
+      return;
+    }
```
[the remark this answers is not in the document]
samuelmaddock: I was also wondering about this. When I run the script, and double-click the .app, it doesn't seem to show me the popup about it being downloaded from the internet 🤔

src/e-pr.js
```diff
@@ -188,5 +188,143 @@ program
...
+    if (!(await fs.promises.stat(distZipPath).catch(() => false))) {
+      fatal(`dist.zip not found within the extracted artifact.`);
+      return;
+    }
+
+    // Extract dist.zip
+    await extractZip(distZipPath, { dir: prDir });
+
+    // Check if Electron.app exists within the extracted dist.zip
+    const electronAppPath = path.join(prDir, 'Electron.app');
```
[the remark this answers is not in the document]
samuelmaddock: Good catch! [705163f](https://github.com/electron/build-tools/pull/679/commits/705163f17ce1882be00150f6232b6fa7cb6a99f3)

src/e-pr.js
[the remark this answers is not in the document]
itsananderson: Maybe when I manually downloaded the artifact, that attribute was added by Chrome to the `.zip` and then propagated to the `.app` when I extracted the zip. If downloading it via a script doesn't encounter the same problem, then all the better :)

itsananderson (approved):

dsanders11 (commented):
- Can we implement this without it being a breaking change to `e pr`? I know we don't guarantee CLI stability, but it's nice to avoid breaking it in case others are using it (such as other tooling like the Electron Build Tools VS Code extension)
- This won't support Windows until we migrate from AppVeyor to GHA (should be soon), might be worth calling out

samuelmaddock:
> Can we implement this without it being a breaking change to `e pr`? I know we don't guarantee CLI stability, but it's nice to avoid breaking it in case others are using it (such as other tooling like the Electron Build Tools VS Code extension)

Agreed. I was able to find a way to do this in 9e2ad2696b201633cd5ce4fb2dc429a1eecad91a 👍 

> This won't support Windows until we migrate from AppVeyor to GHA (should be soon), might be worth calling out

For now this will fail when it can't the artifacts.

```
$ e pr download-dist 44411 --platform=win32 --arch=x64
Failed to find artifact: generated_artifacts_win32_x64
```

src/e
```diff
@@ -169,7 +169,7 @@ program
   .command('backport [pr]', 'Assists with manual backport processes')
   .command('show <subcommand>', 'Show info about the current build config')
   .command('test [specRunnerArgs...]', `Run Electron's spec runner`)
-  .command('pr [options]', 'Open a GitHub URL where you can PR your changes')
+  .command('pr [subcommand]', 'Open a GitHub URL where you can PR your changes')
```
dsanders11, line 172:
```suggestion
  .command('pr [subcommand]', 'Work with PRs to electron/electron')
```

Alternative wordings welcome, but we should change this to reflect that there's more commands here now.

src/e-pr.js
```diff
@@ -188,5 +189,187 @@ program
...
+        owner: 'electron',
+        repo: 'electron',
+        branch: pullRequest.head.ref,
+        name: 'Build',
+        event: 'pull_request',
+        status: 'completed',
+        per_page: 10,
+        sort: 'created',
+        direction: 'desc',
+      });
```
dsanders11, line 246: Will this work with PRs from forks?

dsanders11 (commented):
I'm a +1 on the general concept, but a couple of thoughts on the implementation:

* Would be nice if there was a built-in way to clean out downloaded artifacts so they don't build up over time.
* I think some guard rails around this would be good, to prevent a user from fat fingering a PR number and potentially running code they didn't intend to on their local machine. Perhaps a confirmation prompt that displays the title of the PR, the author, indicates if it's a fork, etc?

package.json
```diff
@@ -28,6 +28,7 @@
     "command-exists": "^1.2.8",
     "commander": "^9.0.0",
     "debug": "^4.3.1",
+    "extract-zip": "^2.0.1",
```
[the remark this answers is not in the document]
samuelmaddock: This is validating actually 👍 I would have reached for that package first, but saw we were using `extract-zip` in many other places so I was aiming for consistency.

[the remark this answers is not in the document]
dsanders11: Yea, historically we've used it but probably time to move on given it's unmaintained. I have loose plans to migrate our other usages of it as appropriate.

samuelmaddock:
@dsanders11 here's sample output from the confirmation prompt I added

```
$ e pr download-dist 44598            
? You are about to download artifacts from:

“feat: add query-session-end and improve session-end events on Windows (#44598)” by savely-krasovsky
https://github.com/savely-krasovsky/electron (fork)

Proceed? (Y/n)
```

```
$ e pr download-dist 44411            
? You are about to download artifacts from:

“feat: service worker preload scripts (#44411)” by samuelmaddock
https://github.com/electron/electron

Proceed? (Y/n)
```

I'm still thinking about how I want to approach cleaning up. Whether it should be another command or automatically checked each time any build-tools command is issued.

src/e-pr.js
samuelmaddock: It does!

yarn.lock
```diff
@@ -4782,14 +4778,7 @@ string_decoder@^1.1.1:
...
 
-"strip-ansi-cjs@npm:strip-ansi@^6.0.1":
-  version "6.0.1"
-  resolved "https://registry.yarnpkg.com/strip-ansi/-/strip-ansi-6.0.1.tgz#9e26c63d30f53443e9489495b2105d37b67a85d9"
-  integrity sha512-Y38VPSHcqkFrCpFnQ9vuSXmquuv5oXOKpGeT6aGrr3o3Gc9AlVa6JBfUSOCnbxGGZF+/0ooI7KrPuUSztUdU5A==
-  dependencies:
-    ansi-regex "^5.0.1"
-
-strip-ansi@^6.0.0, strip-ansi@^6.0.1:
+"strip-ansi-cjs@npm:strip-ansi@^6.0.1", strip-ansi@^6.0.0, strip-ansi@^6.0.1:
```
dsanders11, line 4781: I think you're using an older version of Yarn so these are getting folded back into one, which can actually cause some weird issues (not sure it will in this specific case). If I revert these changes and run `yarn install` with v1.22.22 the only change I get is the `adm-zip` addition.

samuelmaddock:
Opened https://github.com/electron/build-tools/pull/680 for cleaning up artifacts.

ckerr (approved):

samuelmaddock:
Bad news regarding `adm-zip`, it doesn't support symlinks (https://github.com/cthackers/adm-zip/issues/455) which Electron.app contains once extracted. I may have to switch back to `extract-zip` for this reason.

dsanders11:
@samuelmaddock, ah, unfortunate. Thanks for the update, makes sense.

src/e-pr.js
```diff
@@ -134,27 +145,23 @@ function pullRequestSource(source) {
...
-    'Where the changes are going to',
-    guessPRTarget(current()),
-  )
+  .option('-s, --source [source_branch]', 'Where the changes are coming from')
+  .option('-t, --target [target_branch]', 'Where the changes are going to')
   .option('-b, --backport <pull_request>', 'Pull request being backported')
   .action(async (options) => {
-    if (!options.source) {
+    const source = options.source || guessPRSource(current());
+    const target = options.target || guessPRTarget(current());
```
dsanders11, line 155: This change loses the nice help text showing the default. Is there an underlying reason which necessitates this change?

Previous, for example:

```
Options:
  -s, --source <source_branch>   Where the changes are coming from (default: "roller/chromium/main")
  -t, --target <target_branch>   Where the changes are going to (default: "main")
  ```

File: src/e-pr.js [hunk 8]
```diff
@@ -188,5 +195,231 @@ program
...
+      await fs.promises.rm(tempDir, { recursive: true });
+
+      console.log(`${color.success} Downloaded to ${outputDir}`);
+    } catch (error) {
+      // Cleanup temporary files
+      try {
+        await fs.promises.rm(tempDir, { recursive: true });
+      } catch {
+        // ignore
+      }
```
dsanders11, line 419: I think this can just go in a `finally` block rather than repeating it in both blocks?

src/e-pr.js
```diff
@@ -188,5 +195,231 @@ program
...
+          type: 'confirm',
+          name: 'proceed',
+          message: `You are about to download artifacts from:
+
+“${pullRequest.title} (#${pullRequest.number})” by ${pullRequest.user.login}
+${pullRequest.head.repo.html_url}${isElectronRepo ? ' (fork)' : ''}
+
+Proceed?`,
+        },
+      ]);
```
dsanders11, line 259: Personally I think I'd prefer this default to false so user's can't fat finger through, but it's not a sticking point.

dsanders11 (changes requested):
Blocking on the lockfile change regarding `wrap-ansi` and friends.

As another safety mechanism I think it would be good to not allow downloading dist for closed PRs. An option could allow it if we think that's a valid use case, but if someone puts up a malicious PR I don't want it to be easy to download the artifacts after we close it.

It's also not clear to me how this PR handles changes to PRs. I think the short SHA should be worked into the artifact names to avoid confusion? There's currently no indicator which commit the artifacts are for, which seems like a recipe for getting confused about which changes are running.

yarn.lock
samuelmaddock: You're correct, I was on v1.22.17 🧠 
I'll update and fix.

src/e-pr.js
samuelmaddock: The issue is that `fatal()` will call process.exit(1)` and the cleanup won't happen. I'll refactor this to use `process.exitCode = 1`

src/e-pr.js
samuelmaddock: Reasonable preference, I'll add it 👍

src/e-pr.js
samuelmaddock: This seems to only work now when running `e pr open --help`, likely due to the way commander's `isDefault: true` behavior works.

samuelmaddock:
@dsanders11 how would you feel about a prompt like this regarding the closed PR issue

```
? You are about to download artifacts from:

“feat: service worker preload scripts (#44411)” by samuelmaddock
https://github.com/electron/electron

❗❗❗ The pull request is closed, only proceed if you trust the source ❗❗❗

Proceed? (y/N)
```

src/e-pr.js
dsanders11: Gotcha. Unfortunate but OK.

dsanders11:
@samuelmaddock, with that and the default being N, I think I'm good with that. 👍

src/e-pr.js
```diff
@@ -188,5 +195,230 @@ program
...
+        throw new Error(`dist.zip not found in build artifact.`);
+      }
+
+      // Extract dist.zip
+      // NOTE: 'extract-zip' is used as it correctly extracts symlinks.
+      d('unzipping dist.zip to %s', outputDir);
+      await extractZip(distZipPath, { dir: outputDir });
+
+      const platformExecutables = {
+        win32: 'electron.exe',
```
VerteDinde, line 395: As a heads up, we're looking at a PR in e/e right now to move Windows CI over to GitHub actions that might break this (I think we're changing `win32` to `win` to align with GN upstream: https://github.com/electron/electron/pull/44136/commits/510b87c9cfc2b3e955d110e6bcb8ea1705aef231), but I don't think we need to hold this PR on that change, we can update it after the e/e PR lands 🙂

VerteDinde (approved):
Agreed with the proposed changes, this looks good to me now - calling out one potential e/e change that may affect this, but given that PR hasn't merged, I don't think we need to block on it unless we'd like to wait for it to land.

yarn.lock
dsanders11: @samuelmaddock, once you update this I'm good to go on merging. 👍

jkleinsc (changes requested):
For consistency sake in https://github.com/electron/electron/pull/44136, windows artifacts are going to use `win` instead of `win32`, so this PR will need to adjust the Windows artifacts names from process.platform to `win`.

Artifact names will be:

generated_artifacts_win_arm64
generated_artifacts_win_x64
generated_artifacts_win_x86
(eg see https://github.com/electron/electron/actions/runs/12286319546)

samuelmaddock:
@dsanders11 I've updated the artifact download directory to include the short commit hash ([9884569](https://github.com/electron/build-tools/pull/679/commits/9884569f1a3b474c978e0c7b6287b117e1ac36dc)):
`~/.electron_build_tools/artifacts/pr-{number}_{commithash}_{platform}-{arch}`

@VerteDinde @jkleinsc I've replaced win32 with win when downloading artifacts. [47acadb](https://github.com/electron/build-tools/pull/679/commits/47acadb2db919c055169e16fdce2d249c64b9d66)

jkleinsc (approved):

dsanders11 (approved):

PR #694, opened 2025-02-04
chore: remove ci command

codebytere:
As in title. This can essentially all be done with `gh workflow` now, so I don't think there's a need for all this extra logic.


ckerr (approved):

dsanders11 (approved):
🫡

PR #697, opened 2025-02-14
feat: reconstruct-chromium-version command

dsanders11:
Also goes by its less wordy name, `e rcv`. Aims to automate the process of bisecting into a Chromium roll PR by "reconstructing" an intermediate Chromium version of your choice via cherry-picking the relevant PR commits (thanks to our consistent style of including the relevant CL in our commit messages).

e rcv example

### DIFF
--- a/src/e-cherry-pick.js
+++ b/src/e-cherry-pick.js
@@ -8,6 +8,7 @@ const path = require('path');
 const os = require('os');
 const { Octokit } = require('@octokit/rest');
 
+const { getGerritPatchDetailsFromURL } = require('./utils/gerrit');
 const { getGitHubAuthToken } = require('./utils/github-auth');
 const { fatal, color } = require('./utils/logging');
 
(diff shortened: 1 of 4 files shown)


src/e-cherry-pick.js
```diff
@@ -8,7 +8,6 @@ const path = require('path');
 const os = require('os');
 const { Octokit } = require('@octokit/rest');
 
-const { getCveForBugNr } = require('./utils/crbug');
 const { getGitHubAuthToken } = require('./utils/github-auth');
```
jkleinsc, line 11:
```suggestion
const { getGerritPatchDetailsFromURL } = require('./utils/gerrit');
const { getGitHubAuthToken } = require('./utils/github-auth');
```

jkleinsc (changes requested):
Looks like the code removed from src/e-cherry-pick.js is still used, so a require to the newly created utils/gerrit should be added.

src/e-cherry-pick.js
dsanders11: Good catch! Forgot this during the refactor.

jkleinsc (approved):

PR #698, opened 2025-02-18
build(dev-deps): update vitest to 3.0.6

dsanders11:
As in title.


vitest.config.js
```diff
@@ -7,5 +7,6 @@ export default defineConfig({
         singleFork: true,
       },
     },
+    testTimeout: 180_000,
```
dsanders11, line 10: Seems like previously Vitest wasn't enforcing the timeout (which defaults to 5s). 🤔 AFAICT there's no actual difference in test time.

jkleinsc (approved):

PR #733, opened 2025-07-01
feat: rewrite in typescript

MarshallOfSound:
This is chonky so I'm putting it up early.

TODO:
* Tests (mjs -> ts)
* `@electron/build-tools-installer` story (need to add a shim in `src/e` to route to `dist/e`, or update build-tools-installer to point at the new dist folder)


dsanders11 (commented):
LGTM, but we'll need to rebase on `main` as a few PRs have landed since this was written.

PR #761, opened 2025-10-06
chore: Update build-tools to Yarn 4

itsananderson:
This PR updates build-tools to use Yarn 4, following the instructions from @MarshallOfSound

Since the `build-tools` repo is a bit different from others in the Electron ecosystem, testing these changes is slightly more challenging.

I pushed a separate `yarn-4-auto-update` [branch](https://github.com/itsananderson/build-tools/tree/yarn-4-auto-update) to my fork, which modified the auto-update functionality to pull from that same `yarn-4-auto-update` branch. That way I could test the auto-update functionality and make sure `yarn --immutable` is invoked correctly.

Testing the initial installation flow w/ `npm i -g @electron/build-tools` is more tricky, since that package automatically pulls the `main` branch. If anyone has ideas about how to test this, I'd be happy to do that additional testing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "lint-staged": "^15.2.10",
     "nyc": "^17.1.0",
     "prettier": "^3.3.3",
-    "vitest": "^3.0.6"
+    "vitest": "~3.0.6"
   },
   "lint-staged": {
     "*.js": [
--- a/yarn.lock
+++ b/yarn.lock
@@ -1041,16 +1041,6 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@types/chai@npm:^5.2.2":
-  version: 5.2.3
-  resolution: "@types/chai@npm:5.2.3"
-  dependencies:
-    "@types/deep-eql": "npm:*"
-    assertion-error: "npm:^2.0.1"
-  checksum: 10c0/e0ef1de3b6f8045a5e473e867c8565788c444271409d155588504840ad1a53611011f85072188c2833941189400228c1745d78323dac13fcede9c2b28bacfb2f
-  languageName: node
-  linkType: hard
-
 "@types/color-name@npm:^1.1.1":
   version: 1.1.1
   resolution: "@types/color-name@npm:1.1.1"
@@ -1064,13 +1054,6 @@ __metadata:
   dependencies:
     "@types/ms": "npm:*"
   checksum: 10c0/5dcd465edbb5a7f226e9a5efd1f399c6172407ef5840686b73e3608ce135eeca54ae8037dcd9f16bdb2768ac74925b820a8b9ecc588a58ca09eca6acabe33e2f
-  languageName: node
-  linkType: hard
-
-"@types/deep-eql@npm:*":
-  version: 4.0.2
-  resolution: "@types/deep-eql@npm:4.0.2"
-  checksum: 10c0/bf3f811843117900d7084b9d0c852da9a044d12eb40e6de73b552598a6843c21291a8a381b0532644574beecd5e3491c5ff3a0365ab86b15d59862c025384844
   languageName: node
   linkType: hard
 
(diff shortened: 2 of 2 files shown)


src/e-backport.js
```diff
@@ -91,7 +91,13 @@ program
       `Failed to checkout new branch "${manualBpBranch}"`,
     );
 
-    spawnSync(config, 'yarn', ['install'], gitOpts, `Failed to do "yarn install" on new branch`);
+    spawnSync(
+      config,
+      'yarn',
+      ['install', '--immutable'],
```
MarshallOfSound, line 97: I think this tries to use `--immutable` on the electron repo which won't have yarn@v4 yet?

src/e-rcv.js
```diff
@@ -257,7 +257,13 @@ program
       `Failed to checkout new branch "${rcvBranch}"`,
     );
 
-    spawnSync(config, 'yarn', ['install'], spawnOpts, 'Failed to do "yarn install" on new branch');
+    spawnSync(
+      config,
+      'yarn',
+      ['install', '--immutable'],
```
MarshallOfSound, line 263: Same here

package.json
```diff
@@ -57,5 +57,5 @@
       "prettier --write"
     ]
   },
-  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
+  "packageManager": "yarn@4.10.3"
```
MarshallOfSound, line 60: Hm, somehow we lost the SHA here? Running `corepack up` should put the SHA in place

src/e-auto-update.js
```diff
@@ -122,8 +122,8 @@ function checkForUpdates() {
     if (headBefore === git(headCmd)) {
       console.log('build-tools is up-to-date');
     } else {
-      console.log(color.childExec('npx', ['yarn', '--prod'], execOpts));
-      cp.execSync('npx yarn --prod', execOpts);
+      console.log(color.childExec('npx', ['yarn', '--immutable'], execOpts));
+      cp.execSync('npx yarn --immutable', execOpts);
```
MarshallOfSound, line 126: `npx` is unsafe and now we have the yarn cjs file we can just do `node .yarn/releases/yarn-4.10.3.cjs` instead.

I'm also fairly sure this will fail on the first run through because `yarn --prod` will run from the _old_ script while having the new code locally

src/e-backport.js
itsananderson: Ah right 🤦 I will remove these

src/e-auto-update.js
MarshallOfSound: I think we should be OK to just Let This Happen.

We can post in #pineapple internally that this will happen, it will auto-fix on the next update. People won't be broken, they'll get an error _once_ and then it will fix on the next `e` command they run

src/e-auto-update.js
```diff
@@ -122,8 +123,8 @@ function checkForUpdates() {
     if (headBefore === git(headCmd)) {
       console.log('build-tools is up-to-date');
     } else {
-      console.log(color.childExec('npx', ['yarn', '--prod'], execOpts));
-      cp.execSync('npx yarn --prod', execOpts);
+      console.log(color.childExec('node', [yarnPath, '--immutable'], execOpts));
+      cp.spawnSync('node', [yarnPath, '--immutable'], execOpts);
```
MarshallOfSound:
```suggestion
      console.log(color.childExec(process.execPath, [yarnPath, '--immutable'], execOpts));
      cp.spawnSync(process.execPath, [yarnPath, '--immutable'], execOpts);
```

itsananderson:
I've rebased. I'm struggling to reproduce the test timeout errors that are happening in CI, though 🤔

itsananderson:
Ah, so regenerating the `yarn.lock` file bumped `vitest` from 3.0.6 to 3.2.4, and according to this [vitest discussion](https://github.com/vitest-dev/vitest/discussions/6511) this was a regression introduced somewhere between 3.0.9 and 3.1.3, which the maintainers aren't interested in backporting to vitest 3.

From those details, it sounds like the best options are probably to either:

1. Upgrade to vitest 4 to get the fix
2. Pin vitest to `3.0.6` so that this Yarn upgrade doesn't introduce the regression

codebytere (approved):

PR #764, opened 2025-10-21
feat: add --commit-updates option to `e patches`

dsanders11:
When doing Chromium rolls, our standard practice is to bundle all non-content changes to patches (i.e. indexes and line number changes) into a single commit with the message "chore: update patches", and then associate content changes with the upstream CL in individual commits. Doing this by hand is a bit tedious and can accidentally mix content changes into the update patches PR, so let's make it an easy automated process when exporting the patches.

### DIFF
--- a/src/e-patches.js
+++ b/src/e-patches.js
@@ -2,10 +2,11 @@
 
 const fs = require('fs');
 const path = require('path');
+const os = require('os');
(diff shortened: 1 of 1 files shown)


codebytere (approved):
what a delightful PR for me personally 😁

PR #767, opened 2025-10-22
fix: migrate reclient users to siso

codebytere:
Reclient is disabled - everyone should be using either `none` or `siso` at this point.


dsanders11 (approved):
@codebytere, in my testing `use_remoteexec = true` is still needed or the build won't work. I pushed some commits which back out that part of the change, also change the `e-init` defaults, and update the tests. 👍

jkleinsc (approved):

PR #768, opened 2025-10-27
fix: use siso for chrome target instead of reclient

mitchchn:


codebytere (approved):

ckerr:
Is there anything this PR needs now?

PR #845, opened 2026-04-17
feat: `e env <command> [args...]`

ckerr:
`$ e env <command> [args...]` runs any command with build-tools' environment variables set.

This is useful if you wan to use build-tools' env for command-line Electron scripts that require `CHROMIUM_BULDTOOLS_PATH` to be set and for `PATH` to include depot_tools, such as `git commit` and `yarn run lint`.

Related: https://github.com/electron/electron/pull/50858

Related: https://github.com/electron/electron/pull/51034

### DIFF
--- a/README.md
+++ b/README.md
@@ -465,6 +465,20 @@ $ e shell
 # Running "/bin/zsh"
 ```
 
+### `e env <command> [args...]`
+
+`e env` runs any command with the active build-tools environment variables applied.
+
+This is useful for using build-tools' configuration for command-line Electron scripts
+that require `CHROMIUM_BULDTOOLS_PATH` to be set and for `PATH` to include depot_tools,
+such as `git commit` and `yarn run lint`.
+
+```sh
+$ e env git commit
+$ e env yarn run lint
+$ e env mycommand.sh arg1 arg2
+```
+
 ### `e backport <PR>`
 
 `e backport <PR>` assists with manual backport processes on the specified pull request.
--- a/src/_e
+++ b/src/_e
@@ -23,7 +23,7 @@ _e () {
   local line
 
   _arguments -C \
-    "1: :(build debug help init make new node patches pr run show start sync test use)"
+    "1: :(build debug env help init make new node patches pr run show start sync test use)"\
     "*::arg:->args"
 
   local completion_func="_e__${line[1]}"
--- a/src/e.ts
+++ b/src/e.ts
@@ -2,6 +2,7 @@
 
 import * as cp from 'node:child_process';
 import * as fs from 'node:fs';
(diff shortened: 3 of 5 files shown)


dsanders11:
@ckerr, doesn't this already exist in `e depot-tools`, AKA `e d`?

Regarding the ref'd PRs, the goal is to avoid requiring the developer to explicitly run their commands through `e d` or `e env` as this PR adds, and instead implicitly find `build-tools` if they have it, otherwise fall back to looking for `depot_tools` on their path. Historically in `e/e` we've supported building entirely without `build-tools`, hence the fallback behavior I've been adding.

ckerr:
> @ckerr, doesn't this already exist in e depot-tools, AKA e d?

I thought that was only for running depot-tools commands -- the help message says it's for running commands in the depot-tools directory. It works for anything though? If so then this PR is moot.

> Regarding the ref'd PRs, the goal is to avoid requiring the developer to explicitly run their commands through e d or e env as this PR adds, and instead implicitly find build-tools if they have it, otherwise fall back to looking for depot_tools on their path. Historically in e/e we've supported building entirely without build-tools, hence the fallback behavior I've been adding.

Well this is what I've been  hitting all day:

```sh
$ echo $PWD
/home/charles/electron/gn/main/src/electron
$ yarn run lint
$ Error: Couldn't find depot_tools, ensure it's on your PATH
    at getDepotToolsEnv (/home/charles/electron/gn/main/src/electron/script/lib/utils.js:194:11)
    at Object.run (/home/charles/electron/gn/main/src/electron/script/lint.js:120:12)
    at main (/home/charles/electron/gn/main/src/electron/script/lint.js:610:20)
$ export PATH="$PATH:$HOME/electron/build-tools/third_party/depot_tools"
$ yarn run lint
Error: CHROMIUM_BUILDTOOLS_PATH environment variable must be set
    at getDepotToolsEnv (/home/charles/electron/gn/main/src/electron/script/lib/utils.js:198:11)
    at Object.run (/home/charles/electron/gn/main/src/electron/script/lint.js:120:12)
    at main (/home/charles/electron/gn/main/src/electron/script/lint.js:610:20)
```

The changes are now forcing me to go through build-tools, which sounds like that's the opposite of what was intended :dizzy_face:

ckerr:
> I thought that was only for running depot-tools commands, ie the help message says it's for running commands in the depot-tools directory. It works for anything though? If so then this PR is moot.

Sure enough, `e d pwd` returns the actual pwd, not the depot-tools directory. :+1: 

Closing this PR as moot.

ckerr:
Moved discussion to e/e @ https://github.com/electron/electron/issues/51132

PR #848, opened 2026-04-18
test: add cherry-pick unit tests for pure helpers

codebytere:
Extract splitPositionalArgs, computeBatchId, and formatPRTitleAndBody from the cherry-pick action handler, export the existing isUrl and commitSubject helpers, and gate program.parse() behind import.meta.main so the module is importable from tests. Add tests/e-cherry-pick.spec.mjs covering all five exports.

### DIFF
--- a/src/e-cherry-pick.ts
+++ b/src/e-cherry-pick.ts
@@ -22,7 +22,7 @@ const ELECTRON_REPO_DATA = {
   repo: 'electron',
 };
 
-interface PatchDetails {
+export interface PatchDetails {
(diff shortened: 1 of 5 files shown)


dsanders11:
@codebytere, looks like CI is unhappy here.

ckerr (commented):
the refactor sounds like a good idea, needs another update to get the tests passing

dsanders11 (approved):

jkleinsc (approved):

PR #850, opened 2026-04-18
feat: add `--config=<name>` to override active config for one invocation

MarshallOfSound:
Adds a top-level `--config=<name>` flag so any `e` command can be run against a specific build config without changing the active one (no `e use <a>` / `e use <b>` dance).

```sh
e --config=my-release build
e --config=main-debug show root
```

#### How

`e.ts` intercepts `--config` (first arg only, `=` or space form), strips it from argv, and exports `EVM_CURRENT`. `evm-config.ts`'s `getCurrentFileName()` checks that env var before reading `evm-current.txt`. Because Commander spawns executable subcommands as child processes that inherit env, the override reaches every command with no per-subcommand changes.

#### Tests

- [x] New `e-show` spec: two configs, verifies `--config` overrides for one call and leaves persistent current untouched (both `=` and space forms)
- [x] `yarn build` / `yarn lint` / `yarn test` (64/64)

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,14 @@ have multiple build configurations and manage them in a way similar to nvm:
 | nvm current          | e show current     | Show which configuration is currently in use   |
 | nvm use &lt;name&gt; | e use &lt;name&gt; | Change which configuration is currently in use |
 
+To run a single command against a config other than the current one without
+switching, pass `--config=<name>` as the first argument:
+
+```sh
+$ e --config=my-release build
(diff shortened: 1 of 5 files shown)


dsanders11 (approved):

dsanders11:
This yield some weird semantics in some cases, like `e --config=foo use bar`, but don't have a better suggestion so LGTM.

codebytere (approved):

PR #851, opened 2026-04-21
feat: add option to skip gn gen

ckerr:
Add an option to skip `gn gen` when running `e build`.

Rationale: I have a downstream fork of Electron that has a couple of different build configuration and doesn't fit cleanly into e/e's upstream model.  So it has its own script for running `gn gen` with the local setup.

For the most part, I'd like to be using build-tools.  But `e build` unconditionally runs `gn gen`, so it clobbers my downstream build filles. This PR adds an opt-in cli arg for build-tools to skip `gn gen`.

### DIFF
--- a/src/e-build.ts
+++ b/src/e-build.ts
@@ -48,8 +48,27 @@ async function runGNGen(config: SanitizedConfig): Promise<void> {
   await depot.spawn(config, gnPath, execArgs, execOpts);
 }
 
-async function ensureGNGen(config: SanitizedConfig): Promise<void> {
+type GenMode = 'skip' | 'include' | 'only';
+
+async function ensureGNGen(config: SanitizedConfig, genMode: GenMode): Promise<void> {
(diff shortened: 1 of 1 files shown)


src/e-build.ts
```diff
@@ -112,7 +126,8 @@ interface BuildOptions {
 program
   .arguments('[ninjaArgs...]')
   .description('Build Electron and other targets.')
-  .option('--only-gen', 'Only run `gn gen`', false)
+  .option('--gen <mode>', 'Control when to run `gn gen`: on (default), off, or only', 'on')
+  .option('--only-gen', 'Alias for `--gen only`', false)
```
codebytere: let's ditch this redundant option then?

ckerr: @codebytere done in da75d74

codebytere (approved):

src/e-build.ts
dsanders11: We use the `--only-gen` option in CI, though, so we'll need to update those usages there when we bump to this.

PR #854, opened 2026-04-24
docs: reorganize README and document missing commands

codebytere:
Restructure the README into logical command groups (core workflow, running Electron, inspecting state, working with code, managing configs, infrastructure) and add a table of contents for navigation.

Fill in documentation for commands that were undocumented or only briefly mentioned: worktree, pr open, pr download-dist, cherry-pick, rcv, backport, gh-auth, npm, load-macos-sdk, and the show subcommands depotdir / out --path. Expand option tables for init (sanitizers, mas, target-cpu, remote-build, use-https, fork), build (--gen, --no-remote), sync (--3), test, patches (--list-targets, --commit-updates, all), and open (--print).

Add a configuration file reference (schema fields, validation rules, `extends` inheritance) and an environment variables table covering EVM_*, BUILD_TOOLS_SHA, GN_EXTRA_ARGS, ELECTRON_BUILD_TOOLS_* and the other internally-set variables. Point readers at the zsh completion script under tools/zsh/_e.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,9 +63,9 @@ commands and subcommands. See `e --help` or `e help <cmd>` for more details on a
 `e` also borrows another inspiration from nvm: having multiple configurations that you can switch
 between so that one is the current, active configuration. Many choices go into an Electron build:
 
-* Which [Electron branch](https://github.com/electron/electron/branches) is used (e.g. `main`, `36-x-y`)
-* Which [.gn config file][gn-configs] is imported (e.g. [testing](https://github.com/electron/electron/blob/main/build/args/testing.gn) or [release](https://github.com/electron/electron/blob/main/build/args/release.gn))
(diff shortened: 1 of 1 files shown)


dsanders11 (changes requested):
* Linter is unhappy
* I think the "Environment variables" section should only document ones we actually want to tell users about, right now it documented everything. This is non-exhaustive, but I'd definitely remove:
  * `ELECTRON_OVERRIDE_DIST_PATH` - implementation detail
  * `ELECTRON_RUN_AS_NODE` - implementation detail
  * `CI` - implementation detail - if we wanted this to be public API we should give it a proper name and default the value to `CI`.

dsanders11 (approved):

MarshallOfSound (approved):

PR #855, opened 2026-04-29
chore: drop CHROMIUM_BUILDTOOLS_PATH in favor of a buildtools symlink

codebytere:
depot_tools added a stderr warning in CL:7790518  declaring the env var "highly unsupported and may break gn." 

We were setting it because Electron's gclient solution is named `src/electron`, so depot_tools' auto-detection looks for buildtools at `<root>/src/electron/buildtools` and `<root>/buildtools` — neither exists in our layout (the real dir is `<root>/src/buildtools`).

Add ensureBuildtoolsSymlink() that creates `<root>/buildtools` -> `src/buildtools` (dir symlink on POSIX, junction on Windows). That satisfies depot_tools' supported `<gclient_root>/buildtools` lookup and lets us drop the env var entirely.

Wire the helper into `e init`, `e sync`, `e worktree add`, and sanitizeConfig so new and existing checkouts are migrated transparently. sanitizeConfig also strips CHROMIUM_BUILDTOOLS_PATH from saved configs on load. The schema keeps it as an optional field for backwards compatibility but no longer requires it.

### DIFF
--- a/evm-config.schema.json
+++ b/evm-config.schema.json
@@ -135,9 +135,6 @@
           "description": "Path to use as git cache for gclient"
         }
       },
-      "required": [
-        "CHROMIUM_BUILDTOOLS_PATH"
-      ],
       "additionalProperties": {
         "type": "string"
       },
--- a/example-configs/evm.base.yml
+++ b/example-configs/evm.base.yml
@@ -16,5 +16,4 @@ gen:
   out: Testing
 env:
   GIT_CACHE_PATH: /Users/user_name/.git_cache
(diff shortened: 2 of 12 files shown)


jkleinsc (changes requested):
Looks like this change is causing a failure on Windows

dsanders11 (changes requested):
Leaving this open for now in case it doesn't work out, but I think #859 is the approach we want to take to ensure maximum compatibility.

dsanders11:
Closing as https://chromium-review.googlesource.com/c/chromium/tools/depot_tools/+/7813342 landing upstream means we don't need a symlink anymore.

PR #863, opened 2026-05-16
feat: add an option to checkout Electron from a fork

lucasterra:
# Problem

Right now it's impossible to use `@electron/build-tools` to build from a fork. I actively maintain a fork of Electron with features that wouldn't make sense upstream, but I still regularly merge upstream to keep up with Chromium/Node upgrades.

Currently my workaround is to do this:
```bash
# Initialize
e init \
    --root ./electron \
    --import release \
    --remote-build "none" \
    --target-cpu "x64" \
    --fork cool-fork/electron \
    release-x64

# Replace origin with fork in `.gclient`
sed -i 's|https://github.com/electron/electron|https://github.com/cool-fork/electron|g' ./electron/.gclient

# Sync and build
e sync
e build electron:electron_dist_zip --no-remote
```

# Solution

This PR adds a flag for `--fork-as-origin` to `e init` for users who want to checkout Electron directly from their fork rather than `electron/electron`.

This would simplify my build script to this:
```bash
# Initialize, sync and build
e init \
    --root ./electron \
    --import release \
    --remote-build "none" \
    --target-cpu "x64" \
    --fork cool-fork/electron \
    --fork-as-origin \ # 👈 New flag
    release-x64

# Sync and build
e sync
e build electron:electron_dist_zip --no-remote
```

### Behavior

- The existing behavior from `--fork` is maintained:
  - `origin` and `.gclient` checkout from `electron/electron`
  - `fork` remote is pointed to the option passed by `--fork`
- When passing `--fork cool-fork/electron --fork-as-origin`:
  - `origin` and `.gclient` check out from the fork
  - `upstream` is added pointing at `electron/electron`

#### Testing

Added new tests for existing behavior, checking out with SSH, HTTPS and checking out with `--fork-as-origin`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,12 @@ existing root to share it between build configs.
 | `--remote-build <target>`    | Remote-execution backend: `siso` (default) or `none`                             |
 | `--use-https`                | Set git remotes with `https://` URLs instead of `git@github.com:`                |
 | `--fork <user/electron>`     | Add a remote named `fork` pointing at the given GitHub fork                      |
+| `--fork-as-origin`           | Use `--fork` as the checkout `origin` and add `electron/electron` as `upstream`  |
+
+By default, `.gclient` checks out Electron from `https://github.com/electron/electron`. `--fork`
+adds a separate Git remote named `fork` and leaves `origin` pointing at upstream Electron. Use
+`--fork-as-origin` with `--fork` when the checkout source itself should be your fork; in that mode,
+`origin` points at the fork and `upstream` points at `electron/electron`.
 
 **Example**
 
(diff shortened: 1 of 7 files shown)


dsanders11:
> Right now it's impossible to use `@electron/build-tools` to build from a fork.

Can you describe more of the difficulty you're having? There's definitely folks building from a fork using `@electron/build-tools`. You'd generally just switch to the target branch of your fork in your checkout after doing the initial `sync`, and then running `sync` again as needed.

It's a bit inefficient since you'll have the extra sync, but it's the same inefficiency as setting things up to build an older branch of Electron as well.

Alternatively for you, instead of your current workaround you could try `e sync --spec=...` similar to [what we do here](https://github.com/electron/electron/blob/dcb4bef70c701360939eef9a4f1c4e354012e0f4/.github/workflows/pipeline-electron-lint.yml#L72).

lucasterra:
It’s exactly that. The sync → switch → sync causes the pain. The initial sync from electron/main is completely wasted time if I don’t plan on building from main.

And I have to build it 5x for every release (all platforms I support).

The --spec param might work. I’ll try and get back to you. Thanks for the tip.

PR #866, opened 2026-05-22
fix: properly quote Python command in depot_tools check

dsanders11:
The check to see if we have a new enough `depot_tools` to not need `CHROMIUM_BUILDTOOLS_PATH` wasn't working on Windows because the Python command needs to be properly quoted.

### DIFF
--- a/src/utils/depot-tools.ts
+++ b/src/utils/depot-tools.ts
@@ -278,12 +278,15 @@ export function getChromiumBuildtoolsPath(config: ConfigLike): string | undefine
   const disableLogging = process.env['ELECTRON_DEPOT_TOOLS_DISABLE_LOG'];
   process.env['ELECTRON_DEPOT_TOOLS_DISABLE_LOG'] = '1';
 
+  const pythonCommand =
+    "from gclient_paths import GetBuildtoolsPath; print(GetBuildtoolsPath() or '')";
+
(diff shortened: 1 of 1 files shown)


src/utils/depot-tools.ts
```diff
@@ -278,12 +278,15 @@ export function getChromiumBuildtoolsPath(config: ConfigLike): string | undefine
...
+  const pythonCommand =
+    "from gclient_paths import GetBuildtoolsPath; print(GetBuildtoolsPath() or '')";
+
   let result;
   try {
     result = spawnSync(
       config,
       'python3',
-      ['-c', 'from gclient_paths import GetBuildtoolsPath; print(GetBuildtoolsPath() or "")'],
+      ['-c', process.platform === 'win32' ? `"${pythonCommand}"` : pythonCommand],
```
ckerr, line 289: Is the platform check needed? Seems like it would be less work + safer to always quote it

dsanders11: Not necessary, but I was opting for the least amount of change to avoid unintentionally breaking anything else. 😅 This code is temporary and will be pulled out in a few months anyway, so was opting to only change the already broken code path. Should be fine to do on all platforms.

jkleinsc (approved):

PR #869, opened 2026-05-27
feat: add `e auto-roll`

MarshallOfSound:
> Depends on electron/agent-workflows#11 (**merged**), which roots the roll
> bundle at a fetchable base commit. Without that, the bundle's base is a
> local-only commit and `apply` fails with `Repository lacks these
> prerequisite commits`.

## What

Adds `e auto-roll` — list the Chromium auto-rolls produced by the
[agent-workflows][aw] service, pick one interactively, download its git bundle,
and apply the proposed commits on top of your local `roller/chromium/main`.

```
e auto-roll [--branch <name>] [--remote <name>] [--no-fetch] [--limit <n>] [--run <id>] [--no-apply]
```

- Authenticates with the service's Cloudflare Access (browser SSO the first
  time; token cached under `~/.cloudflared`, reused after).
- Lists recent successful runs, shows the agent's summary + `base → head`, and
  asks before doing anything.
- Fetches the latest roller branch, then applies the roll: fast-forward when the
  branch is exactly at the roll's base, otherwise cherry-picks the commits on
  top. Conflicts stop with the bundle ref left at `refs/auto-roll/<run>` so you
  can resolve and `git cherry-pick --continue`.

## How the Cloudflare Access auth works (no `cloudflared` dependency)

`src/utils/cf-access.ts` ports cloudflared's browser-login token-transfer flow:
discover the Access app via the login redirect, open the login URL, long-poll
`login.cloudflareaccess.org` for the encrypted token, and decrypt it (NaCl box,
via `tweetnacl`). The token is sent in the `Cf-Access-Token` header, and cached
in `~/.cloudflared` so it interops with the real binary if present.

## Notes

- New dependency: `tweetnacl` (audited, zero-dep, no install scripts) for the
  NaCl box decryption — Node's `crypto` has X25519 but not XSalsa20/Poly1305.
- `git`-level apply mechanics (bundle verify, fetch-by-sha, fast-forward vs
  cherry-pick) were validated against synthetic repos locally.

[aw]: https://github.com/electron/agent-workflows

### DIFF
--- a/src/e-auto-roll.ts
+++ b/src/e-auto-roll.ts
@@ -1,0 +1,497 @@
+#!/usr/bin/env node
+
+import * as cp from 'node:child_process';
+import * as fs from 'node:fs';
+import * as os from 'node:os';
+import * as path from 'node:path';
+import * as stream from 'node:stream';
+import { pipeline } from 'node:stream/promises';
+
+import { program } from 'commander';
+import * as inquirer from '@inquirer/prompts';
+import debug from 'debug';
+import extractZip from 'extract-zip';
+
+import * as evmConfig from './evm-config.js';
+import { AccessClient } from './utils/cf-access.js';
+import { progressStream } from './utils/download.js';
+import { color, fatal } from './utils/logging.js';
+
+const d = debug('build-tools:auto-roll');
+
+/** The agent-workflows dashboard/API, fronted by Cloudflare Access. */
+const AGENTS_URL = new URL('https://agents.electronjs.org');
+/** Branch the chromium roller pushes to (and that we replay rolls onto). */
+const DEFAULT_ROLLER_BRANCH = 'roller/chromium/main';
+/** Artifact name prefix the workflow uses for the git bundle. */
+const BUNDLE_ARTIFACT_PREFIX = 'chromium-roll-';
+const BUNDLE_FILE = 'roll.bundle';
+const BUNDLE_META_FILE = 'bundle.meta';
+const ROLL_SUMMARY_FILE = 'roll-summary.md';
+
+interface RunSummary {
+  id: number;
+  run_number: number;
+  status: string;
+  conclusion: string | null;
+  head_sha: string;
+  actor: string | null;
+  html_url: string;
+  created_at: string;
+  updated_at: string;
+  run_started_at: string | null;
+}
+
+interface RunsResponse {
+  workflow_runs: RunSummary[];
+  total_count: number;
+}
+
+interface ArtifactMeta {
+  id: number;
+  name: string;
+  size_in_bytes: number;
+  created_at: string | null;
+  expired: boolean;
+}
+
+interface AutoRollOptions {
+  branch: string;
+  remote: string;
+  fetch: boolean;
+  limit: string;
+  apply: boolean;
+  run?: string;
+}
+
+function shortSha(sha: string): string {
+  return sha.slice(0, 9);
+}
+
+function timeAgo(iso: string | null): string {
+  if (!iso) return 'unknown';
+  const then = new Date(iso).getTime();
+  if (Number.isNaN(then)) return 'unknown';
+  const secs = Math.max(0, Math.round((Date.now() - then) / 1000));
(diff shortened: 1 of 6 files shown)


jkleinsc (approved):

src/utils/cf-access.ts
```diff
@@ -0,0 +1,343 @@
...
+  const timer = setTimeout(() => controller.abort(), POLL_TIMEOUT_MS);
+  let res: Response;
+  try {
+    res = await fetch(url, {
+      headers: { 'User-Agent': USER_AGENT },
+      signal: controller.signal,
+    });
+  } finally {
+    clearTimeout(timer);
+  }
```
dsanders11:
```suggestion
  const res = await fetch(url, {
    headers: { 'User-Agent': USER_AGENT },
    signal: AbortSignal.timeout(POLL_TIMEOUT_MS),
  });
```

Using `AbortSignal.timeout` simplifies this a good amount.

dsanders11: I think there's a bug here, as the intention seems to be that a timeout would abort the request and then the loop in `runTransfer` would start another request, but at the moment `pollTransfer` will throw an abort error that will bubble up out of `runTransfer`.

Additionally, `signal: AbortSignal.timeout(POLL_TIMEOUT_MS)` would simplify this code some.

dsanders11 (changes requested):

src/e-auto-roll.ts
```diff
@@ -0,0 +1,497 @@
...
+  remote: string,
+): void {
+  const bundlePath = path.join(extractDir, BUNDLE_FILE);
+  const metaPath = path.join(extractDir, BUNDLE_META_FILE);
+
+  const meta = fs.existsSync(metaPath) ? parseBundleMeta(fs.readFileSync(metaPath, 'utf8')) : {};
+  if (!meta.base || !meta.head) {
+    fatal(`${BUNDLE_META_FILE} is missing base/head — refusing to apply an unverifiable bundle`);
+  }
+  const { base, head } = meta as { base: string; head: string };
```
dsanders11:
```suggestion
  const { base, head } = meta;
```

Unnecessary type assertion.

src/e-auto-roll.ts
```diff
@@ -0,0 +1,497 @@
...
+  const total = Number(zipRes.headers.get('content-length') ?? 0);
+  const source = stream.Readable.fromWeb(
+    zipRes.body as Parameters<typeof stream.Readable.fromWeb>[0],
+  );
+  const out = fs.createWriteStream(zipPath);
+  if (total > 0 && process.stdout.isTTY) {
+    await pipeline(source, progressStream(total, '  [:bar] :mbRateMB/s :percent :etas'), out);
+  } else {
+    await pipeline(source, out);
+  }
```
dsanders11:
```suggestion
  const out = fs.createWriteStream(zipPath);
  if (total > 0 && process.stdout.isTTY) {
    await pipeline(zipRes.body, progressStream(total, '  [:bar] :mbRateMB/s :percent :etas'), out);
  } else {
    await pipeline(zipRes.body, out);
  }
```

I've tested this locally, you can pass the web stream directly to `pipeline`. This change will need to drop the now unused import on L7 as well.

dsanders11 (changes requested):

src/utils/cf-access.ts
MarshallOfSound: good catch. heads up that `AbortSignal.timeout` on its own still bubbles (just as a `TimeoutError`), so I'm catching it and falling through to the retry loop. fixed 👍

src/e-auto-roll.ts
MarshallOfSound: yep, the guard + `fatal()` narrows it to string anyway. dropped

src/e-auto-roll.ts
MarshallOfSound: oh nice, didn't realise pipeline took the web stream directly. done, dropped the `node:stream` import too 👍

dsanders11 (approved):

PR #908, opened 2026-09-20
build: update @electron/lint-roller to 4.0.0

MarshallOfSound:
`@electron/lint-roller` was still on 1.13, whose `electron-markdownlint` wrapper was removed back in 3.0; current lint-roller only provides the shared markdownlint config and custom rules (and 4.0.0 drops the transitive eslint/standard tree).

`lint:markdown` now runs `markdownlint-cli2` directly with a `.markdownlint-cli2.jsonc` that extends the shared config, loads the custom rules, and carries over the old `.markdownlintignore` entries. The one rule that's new since 1.13's markdownlint and fires here, `table-column-style`, is turned off because the README's aligned tables deliberately let a few long rows overflow.

---
_Generated by [Claude Code](https://claude.ai/code/session_01GaskPz9P37WZvpQ5zKit7V)_


MarshallOfSound:
@SocketSecurity ignore npm/@electron/lint-roller@4.0.0

